PR #8, opened 2020-01-24
Pic 196 case list view

theDustRoom:
Included WireMock and sample data for testing locally and  to support the integration tests.
Included WireMock in CircleCi build process.
Updated BDD tests to include tests for Case list view.
Updated unit tests.
Added nodemon for auto rebuild during development.


views/case-list.njk
```diff
@@ -1,7 +1,73 @@
+{% set tableData = {
+  head: [
+    { text: "Defendant" },
+    { text: "Probation record" },
```
katie-cawthorne: Alistair has changed the content for the "Probation record" column, it's now "Probation status"

theDustRoom: Ack; sorry I'll change that.

PR #12, opened 2020-01-30
Pic 207 pagination

theDustRoom:


app.js
```diff
@@ -39,6 +39,7 @@ app.use('/assets', [
   express.static(path.join(__dirname, '/node_modules/govuk-frontend/govuk/assets')),
   express.static(path.join(__dirname, '/node_modules/@ministryofjustice/frontend/moj/assets'))
 ])
+app.use('/moj', express.static(path.join(__dirname, '/node_modules/@ministryofjustice/frontend/moj')))
```
theDustRoom: I noticed that this hadn't been included

public/stylesheets/app/_overrides.scss
```diff
@@ -5,3 +5,7 @@
 .moj-sub-navigation__link {
   @extend .govuk-link--no-visited-state;
 }
+
+.moj-pagination__results {
+  @extend .govuk-body;
```
theDustRoom: This is broken in the Pagination component so this override is necessary for now.

views/partials/layout.njk
```diff
@@ -86,6 +86,7 @@
 {% endblock %}
 
 {% block bodyEnd %}
+  <script src="/moj/all.js"></script>
```
theDustRoom: I noticed that this hadn't been included

PR #13, opened 2020-01-30
Pic 205: know when case list was last updated

jhevans:
Left a couple of comments about non-obvious things but should be largely self explanatory


app.js
```diff
@@ -16,7 +16,7 @@ const env = nunjucks.configure([
 ], {
   autoescape: true,
   express: app,
-  watch: true
+  watch: process.env.WATCH_TEMPLATES
```
jhevans: This change was done to prevent the unit tests from hanging after completion, the watcher was holding on to a file handle so jest refused to exit. Have removed template watching unless explicitly turned on for this reason as we probably don't want it in prod anyway.

package.json
```diff
@@ -16,6 +16,8 @@
   },
   "scripts": {
     "precommit": "yarn lint && yarn unit-test && yarn int-test",
+    "pre-commit-lite": "yarn lint && yarn unit-test",
```
jhevans: Left existing precommit command in case anyone is using it

app.js
theDustRoom: Yep, good call.

package.json
theDustRoom: LGTM

public/stylesheets/style.scss
```diff
@@ -5,3 +5,12 @@ $govuk-brand-colour: #204060;
 @import "node_modules/govuk-frontend/govuk/all";
 @import "node_modules/@ministryofjustice/frontend/moj/all";
 @import "app/all";
+
+// 'pac' prefix used for custom styles to prevent collisions
+.pac-last-updated {
+    @include govuk-media-query($from: tablet) {
+        top: 12px;
+        float: right;
+        position: relative;
```
theDustRoom: You should clear the float here with:

@include govuk-clearfix;

mappings/case-list.json
```diff
@@ -15,6 +15,7 @@
     },
     "status": 200,
     "jsonBody": {
+      "lastUpdated": "2019-12-03T08:30:00",
```
theDustRoom: Actually, might be better to change this to use the current date:

"lastUpdated": "{{now format='yyyy-MM-dd'}}T08:30:00",

mappings/empty-case-list.json
```diff
@@ -15,6 +15,7 @@
     },
     "status": 200,
     "jsonBody": {
+      "lastUpdated": "2019-12-03T08:30:00",
```
theDustRoom: Actually, might be better to change this to use the current date:

"lastUpdated": "{{now format='yyyy-MM-dd'}}T08:30:00",

mappings/case-list.json
jhevans: Yeah I did consider this but I'm kinda wary as it means both the data and the test will be dynamic and therefore non-deterministic. I agree it's nice to have it show realistic data though, will do as you suggest and we can review it if it causes any problems, which tbf I doubt it will.

jhevans: ...strictly speaking it is deterministic actually coming to think of it so even less of a problem. Slight chance of edge cases causing arbitrary failures but unlikely.

PR #25, opened 2020-03-03
Pic 248 filter component

theDustRoom:
I've created PIC-304 to correctly configure and test the session as it's not really in scope for this work but was needed in order for this to function.

Another thing to note is that we will need to tidy any and all UI component-level JavaScript into a single file and include as part of the build process at some point. In have created  PIC-305 for this work.


README.md
```diff
@@ -40,7 +40,7 @@ If unspecified, the app will use your local court-case-service app but you can a
  
 Start the standalone WireMock server
 ```
-java -jar wiremock-standalone-2.26.0.jar --global-response-templating --port 9091
+java -jar wiremock-standalone-2.26.2.jar --global-response-templating --port 9091
```
jhevans: Could you add instructions on how to start the Redis server etc? There's a few moving parts involved in running  the integration tests now so we ought to document it, would be nice to get it running from one command if possible though that's a bit more involved.

.circleci/config.yml
```diff
@@ -53,7 +53,14 @@ executors:
     working_directory: ~/app
   builder:
     docker:
-      - image: circleci/node:12.14.1-browsers
+      - image: circleci/node:12.16.1-browsers
+    working_directory: ~/app
+  integration-tests:
+    docker:
+      - image: circleci/node:12.16.1-browsers
+      - image: bitnami/redis:5.0
```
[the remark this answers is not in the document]
theDustRoom: Redis is used for session storage as the default MemoryStore is not suitable for production.
We will most likely use Redis for caching later as well.

README.md
theDustRoom: I've updated the README with instructions on how to install Redis.
This will most likely be removed and setup with Docker when we look at this properly as part of PIC-304

public/stylesheets/app/_filter.scss
```diff
@@ -24,7 +24,7 @@
   border-top: 0;
   z-index: 9999;
   max-height: 250px;
-  overflow: scroll;
+  overflow-y: auto;
```
theDustRoom: Corrected the scroll behaviour

views/components/filter/template.njk
```diff
@@ -81,8 +81,8 @@
                     <div class="pac-filters-applied">
                         {% for item in filter.items %}
                             {% if item.checked %}
-                                <div class="moj-filter__tag pac-filter__tag"
-                                     data-controls="checkbox-{{ filter.id }}-{{ loop.index }}">{{ item.label }}</div>
+                                <button class="moj-filter__tag pac-filter__tag" type="button"
```
theDustRoom: Minor accessibility fix

README.md
theDustRoom: Edit. I've included docker-compose which starts a Redis server for simplicity.
I've updated the README accordingly.

PR #28, opened 2020-03-19
PIC-309 Attendance Data and PIC-332 Time Elapsed on Order

theDustRoom:


integration-tests/integration/common/common.spec.js
```diff
@@ -54,6 +58,16 @@ And('I should see the body text {string}', $text => {
   cy.get('.govuk-body').contains($text)
 })
 
+And('I should see the text {string} within element with class {string}', ($text, $class) => {
+  cy.get(`.${$class}`).contains($text)
+})
+
+And('I should see the following elements with {string} class text', ($class, $data) => {
```
theDustRoom: This is similar to the method above but tests for multiple elements in sequence to ensure data is correct within specific groups of text.

The method above is still required for simple tests on pages without a required structure.

PR #35, opened 2020-03-31
Moves scans to nightly scheduled build

robinsillem:


robinsillem:
https://support.snyk.io/hc/en-us/requests/4341

PR #44, opened 2020-04-22
Pic 355 last pre sentence report

katie-cawthorne:
- Updated convictions test data to align with court-case-service response
- Added front end display for last pre-sentence report for current offender
- Added BDD tests

### DIFF
--- a/views/case-summary-record.njk
+++ b/views/case-summary-record.njk
@@ -29,8 +29,8 @@
                 <h2 class="govuk-heading-m">Current orders ({{ activeOrders | length }})</h2>
                 {%- for order in activeOrders %}
                     <p class="govuk-body govuk-!-margin-bottom-0"><a href="record/{{ order.convictionId }}" class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">{{ order.sentence.description }}</a></p>
-                    <p class="govuk-body govuk-!-margin-top-1 govuk-!-margin-bottom-0">{{ order.offences[0].detail.description }}</p>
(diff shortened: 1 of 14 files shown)


views/case-summary-record.njk
```diff
@@ -92,6 +92,13 @@
                 <p class="govuk-body">Not active</p>
             {% endif %}
 
+            {% for order in activeOrders %}
```
theDustRoom, line 95: If there are multiple active orders with pre-sentence reports, this will display multiple last pre-sentence reports in the UI.

We'll need to find the last pre-sentence report from however many there may be and display only that one.

theDustRoom: Actually; I think that there could be a scenario where the last pre-sentence report could actually be in a terminated order.

If they have a long-term order and they re-offend, then they're given a different order that expires before their existing one... then that Pre-Sentence Report would actually be the last one.

theDustRoom: Ah... after speaking with Charan there might actually be no Pre-Sentence Report on a current order but there could be one on a previous order.

So... we do need to run through all orders to find the actual Last Pre-Sentence Report.

katie-cawthorne: I did think about this but then when I looked at the ticket for the community-api implementation, by default it should only return the last found PSR on the probation record checking current events and reports delivered in the last year.

katie-cawthorne: After speaking with Chris the date filter to return the last PSR is now going to be handled in the client, so I'll look at this now 🙂

theDustRoom:
I've updated the code to look through all orders and display the most recent PSR.

@katie-cawthorne Can you have a look through the changes and give everything a check to make sure it's all good - then we can get this merged.

PR #45, opened 2020-04-24
PIC-435 run snyk vulnerability scans on PRs and fail builds for high severity issues

tomsmyers:
This PR contains the following changes:

* run snyk scans on PRs, changes in master, and nightly for both docker and app vulnerabilities.
* fail builds and deployments when high severity issues are found.
* only report vulnerability snapshots for scans 
* refactor the way we share the app version between build stages by using `mem/recall` instead of a local file.


jhevans:
LGTM 👍

PR #50, opened 2020-04-28
Pic 427 align test data

katie-cawthorne:
- Updated mock data to align to court-case-service response.
- Removed embedded 'data' field as per ticket [PIC-376](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=366&projectKey=PIC&modal=detail&selectedIssue=PIC-376)
- Removed use of 'data' field in views, this requires the following to be added to the court-case-service court case response that I have added to ticket [PIC-377](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=366&projectKey=PIC&view=planning&selectedIssue=PIC-377&issueLimit=100):
  - listNo
  - pnc
  - dateOfBirth
  - defendant sex
  - nationality_1
  - nationality_2

- Updated integration tests


integration-tests/integration/features/case-list.feature
```diff
@@ -106,52 +106,13 @@ Feature: Case list
...
     And I should see pagination link "Previous" with href "?page=4"
-    And I should not see pagination link "1"
-    And I should not see pagination link "2"
+    And I should see pagination link "1" with href "?page=1"
+    And I should see pagination link "2" with href "?page=2"
     And I should see pagination link "3" with href "?page=3"
     And I should see pagination link "4" with href "?page=4"
     And I should not see pagination link "5"
     And I should see pagination page "5" highlighted
-    And I should see pagination link "6" with href "?page=6"
```
theDustRoom, line 115: We need more pages in order to test that the pagination works as expected as with 5 pages we only ever see:
1 2 3 4 5

With more pages we should then see the current page in the middle like:
3 4 **_5_** 6 7

And then we ensure that when we're on the last page we'd see:
3 4 5 6 **_7_**

We don't need 207 cases like before but we do need about 140 I think in order to test it fully.

katie-cawthorne: Sure no problem, I did think that may be the case...I'll look at it now.

PR #67, opened 2020-05-22
PIC-479 Swagger Validation POC

astroash:
### Changes:
- install `swagger-client` for js (currently not in use)
- install `swagger-model-validator` for validating mocks to swagger spec
- setup proof of concept test to show how modules interact

### File Structure:
```
tests
    |
    | - - mock_validation
          |
          | - -  POC.test.js
          | - -  swagger-model.json
```

### Description:

This PR shows how to set up `swagger-model-validator` to validate mock data against swagger spec. It is currently using a hardcoded version of the spec, however this can be switched out for a hosted version by using the `swagger-client` module. This would ensure the spec is up to date but leaves the tests dependent on an external resourse.

Test is currently failing as the mocs do not resprensent the same structure. Half of errors are due to js type inference 🙄

**Current issues:**
```js
{
    "errorCount": 5, 
    "errors": [
      [Error: caseId (951609) is not a type of string], 
      [Error: caseNo (668911253) is not a type of string], 
      [Error: courtRoom (6) is not a type of string], 
      [Error: defendantAddress ([object Object]) is not a type of undefined], 
      [Error: sessionStartTime ({{now format='yyyy-MM-dd'}}T09:30:00) is not a type of date-time]
    ]
}
```

### Todo:

- [x]  write test that check swagger spec against hosted using swagger diff
- [ ] test all mocks
- [ ] add step to  replace `wiremock` date values
- [ ] fix errors in mocs

___
_**Side note:** commits follow [gitmoji standard](https://gitmoji.carloscuesta.me/) to add visual referencing for ease of reading commit messages_


jhevans:
Good work 👍 not too problematic to work with then?

Re the type inference issue - I assume that goes away if we update our mocks to use strings across the board? Or is there any ability to do more lenient checking in cases like that? 

Do you know why it's complaining about defendantAddress? It seems to match with the swagger spec from what I can see.

The `sessionStartTime` issue is a bit of a PITA, but also a bit of an edge case. We could potentially run whatever templating engine Wiremock uses against the json to bring back a valid value there or just do a dumb substitution before validation though so not insurmountable I suppose.

What do you think, worth pursuing?

astroash:
> not too problematic to work with then?

I think it's great :) Especially if it's linked up to a live swagger spec, then there's no dev overhall when it updates, and it's a little like integration testing, as the data is coming directly from the backend. However the downside is that it makes those particular tests a bit more fragile.

> Re the type inference issue - I assume that goes away if we update our mocks to use strings across the board? Or is there any ability to do more lenient checking in cases like that?

It's been a long time since I've used `axios` with js rather than ts, I cannot remember if it automatically parses `strings` of `numbers` to `numbers`, either way, we'll need to ensure that the mocks accurately represent what is returned by axios, and if they automatically infer the type then manually configure axios `transformResponse` to handle the values how we want

> The sessionStartTime issue is a bit of a PITA, but also a bit of an edge case. 

Yeah, I wasn't sure what was happening there, whether that was psudo code written in the mocks or that's what it was expecting the value to be. I decided to leave all the mocks as is to show what the current state is

> What do you think, worth pursuing?

Yeah, this seems like a good idea.

However...

An alternative to validating with `swagger` would be to generate the mocks using the swagger spec. This would remove the step of having to generate mock data ourselves. I've had a brief look at libraries that can do this, but I imagine the flow would be something along the lines of:

- before all tests run generate mocks from swagger spec (this can be live hosted or local file). This could be generated and stored during `globalSetup`
- Tests use mocks that are definitely the shape of the api
- Daily schedule for running tests in circleci to alert if changes to the spec have 
  - only works if mocks are generated from a live 
  - gets into issues of different spec for different enviroments (I'm unsure whether you release backend & frontends simultaniously)

_don't quote me on any of this_ :)

theDustRoom:
@astroash The sessionStartTime is set so that WireMock replaces this with the current date so that we're always looking at "today".

jhevans:
🤔 yes that would work. Would that not be quite difficult though? You'd have to automate generation of your mocks as well as your BDD globals.  Also part of the problem we have is that front end dev is outpacing the back end - so in some cases we simply wouldn't have the swagger to hand to build from. 

It occurs to me not having the swagger spec actually presents problems no matter how we do it because it will still fail because the spec isn't there.. so there would have to be a manual intervention at the point when the back end was implemented to start checking the interface. Need to think about how we manage that... but I digress.

I also wonder whether it would be good to keep the BDD tests and the specification checks separate. My reasoning being that it's then immediately obvious when it breaks whether it's an issue with the BDD or an issue with the integration. If you build your mocks from the spec then that could result in broken tests because the interface has changed - or because of a breaking change to the application. 

Not sure which would be better in terms of how we update the spec, requesting it dynamically means you're working with a moving target which could be problematic. Or you could get all fancy and raise a PR automatically on changes to court-case-service to update prepare-a-case with the latest spec. That then gives you an opportunity to fix any issues before it's merged into prepare-a-case master...

I think my brain needs a rest...

astroash:
> I also wonder whether it would be good to keep the BDD tests and the specification checks separate.

I agree. On second thought you're correct, this would be conflating validation of test data and BDD tests together. In the spirit of agile the solution demoed in this PR would be the minimum step to ensure that mock data used for BDD tests is correct.

>Not sure which would be better in terms of how we update the spec, requesting it dynamically means you're working with a moving target which could be problematic. Or you could get all fancy and raise a PR automatically on changes to court-case-service to update prepare-a-case with the latest spec. That then gives you an opportunity to fix any issues before it's merged into prepare-a-case master...

It's been a little while since I've setup github webhooks, but I think you would need to run a little server for this to accept the webhook payload when a PR is merged, then hit the github api to 

Is there a slackbot integration for MoJ github? This may be a more light weight solution to give a reminder that the spec _may have changes_.

Alternatively you could have a prepush githook that runs a [swagger diff](https://swagger.io/blog/api-development/using-swagger-to-detect-breaking-api-changes/)check, then it will be the devs responsiblity to notify the rest of the team that the there are changes to the api.

PR #74, opened 2020-06-05
PIC-534 Prepare a Case - Dockerfile

theDustRoom:
Removed use of Yarn and change to use NPM
Corrected dependency for node-sass-middleware
Optimised production build in Dockerfile


theDustRoom:
Snyk has found vulnerabilities in node-sass but there are no fixes for this as yet.

It's being monitored so will be updated once a fix is in place.

PR #78, opened 2020-06-16
PIC-454 Production Domain. Included Helmet

theDustRoom:
Included :lock: Helmet for security and set maxAge as per ticket.
Also updated dependencies.


theDustRoom:
Removed node-sass-middleware as it's a bit rubbish and is causing all manner of issues with Snyk

PR #79, opened 2020-06-22
PIC-509 Swagger Mock Validation

theDustRoom:
Updated mock validation tests
Updated package.json with additional mock validation tests
Corrected mappings based on validation tests
Updated dependencies

### DIFF
--- a/mappings/case-list-full.json
+++ b/mappings/case-list-full.json
@@ -18,10 +18,10 @@
       "lastUpdated": "{{now format='yyyy-MM-dd'}}T08:30:00",
       "cases": [
         {
-          "caseId": 957781,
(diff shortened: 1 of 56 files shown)


theDustRoom:
I have configured this to test against live swagger docs either locally or on a published service de/pre-prod/prod.

I didn't see the point in pulling a snapshot of the swagger docs to run on CI as this can easily be outdated/incorrect and is essentially a redundant task as it's the same as running the tests locally and not on the CI.

It would be beneficial if the tests could run on CircleCi at least against Dev but we will have to see if this is possible/acceptable.

I have not included tests for personal details or attendance as these endpoints are deprecated in the backend and the frontend needs to be aligned again.

PR #100, opened 2020-07-16
Pic 541 add app insights

katie-cawthorne:
Added Application Insights config
Added secrets in environments

### DIFF
--- a/server/application-version.js
+++ b/server/application-version.js
@@ -1,4 +1,3 @@
-// eslint-disable import/no-unresolved,global-require
 const fs = require('fs')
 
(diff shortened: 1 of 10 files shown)


server/application-version.js
```diff
@@ -0,0 +1,8 @@
+// eslint-disable import/no-unresolved,global-require
```
theDustRoom, line 1: You should take out the `eslint` comments as we don't use it.

katie-cawthorne: Yep sure I'll take out now

PR #117, opened 2020-08-17
Upgrade dependencies

theDustRoom:


app.js
```diff
@@ -52,7 +52,9 @@ module.exports = function createApp ({ signInService, userService }) {
 
   app.set('view engine', 'njk')
 
-  app.use(helmet())
+  app.use(helmet({
+    contentSecurityPolicy: false
```
theDustRoom, line 56: Helmet 3 did not include `Content-Security-Policy` and Helmet 4 does. As the pen test was fine with Helmet 3 settings, I've updated the config to disable this.

PR #120, opened 2020-08-19
Pic 679 frontend alignment with prototype

katie-cawthorne:
Changes made:
1. Underline added to `Sign out` link
2. Removed link from account name
3. Hover and focus updated to match prototype
4. Previous day paging link alignment
5. Divided timestamp into 2 separate elements and added CSS border to separate elements
6. Changed format of months from truncated to full spelling in case list navigation
7. Filter - close dropdown outside of an expanded filter
8. Removed PNC from key details bar of 'no record' and 'possible record' cases
9. Removed DOB from key details bar for all cases
10. Date format changed in Appearance on case summary, removed 'st, th, rd'
11. Removed CPS pack from case summary
12. Renamed 'Personal details from police' to 'Defendant details'
13. Moved Nationality above address
14. Removed aria current role on primary nav item when in case

- Updated BDD tests for all the above changes


theDustRoom (approved):
LGTM

PR #129, opened 2020-09-03
:bug: PIC-690: Fixed filter for court room and probation status

katie-cawthorne:
- Filter broken on prod due to data for courtroom including zero eg, `03`, I have reformatted the courtroom to remove the zero.
- This also includes code for ticket PIC-691 regarding the probation status, the filter is expecting the probation status for `Previously known` to be in that format but instead is `Previously Known` - the K is capitalised, so this had broken the filter for probation status. 
- Updated case-list test data to reflect correct data format for fields `courtRoom` and `probationStatus`
- Updated BDD tests


theDustRoom (approved):
LGTM

PR #130, opened 2020-09-03
PIC-618 Review Manual Match

theDustRoom:


theDustRoom:
Marked this as draft as I'm going to run through this with @sarahyjja tomorrow.

server/routes/index.js
```diff
@@ -271,8 +271,37 @@ module.exports = function Index ({ authenticationMiddleware }) {
...
+    const detailResponse = await getDetails(req.params.crn)
+    templateValues.title = 'Link an nDelius record to the defendant'
+    templateValues.details = {
+      ...detailResponse
+    }
+    templateValues.session = {
+      ...req.session
+    }
+    res.render('match-manual', templateValues)
+  })
```
katie-cawthorne, line 290: There's no mention on the ticket but we probably need to have an error displayed to the user if the crn can't be found, otherwise the user will just see an empty table.

katie-cawthorne (approved):
Looks good and works fine, just the comment about displaying an error to the user if the crn can't be found. However, there's no mention of this on the ticket so happy to approve.

server/routes/index.js
theDustRoom: Yes, you're right. I thought of this but forgot to mention it. I'll raise it with Charan

PR #133, opened 2020-09-07
PIC-725 BUGFIX - Invalid date on matching screen

theDustRoom:
Corrects data to display most recent event on match screen


katie-cawthorne (approved):
👍

PR #134, opened 2020-09-07
Pic 690 fix court room filter

katie-cawthorne:
Small fix...think it's to do with the data for courtRoom coming through with a leading zero.


theDustRoom (approved):
LGTM

PR #135, opened 2020-09-07
PIC-646 Add button for cases with no record

sarahyjja:
- add html button
- css modification
- BDD scenarios
- cy tests
- matching link


katie-cawthorne (approved):
Looks good, nice one ✨

PR #137, opened 2020-09-07
BUGFIX - Possible records and probation status are displayed together

theDustRoom:
Includes CRN lookup and error message that was missed on PIC-647 (sorry)
Removed duplicate scenarios from BDD resulting from merge
Made 1st page of case list data more consistent with real data (no longer shows breach and SSO for defendants without a current order)


katie-cawthorne (approved):
LGTM

PR #138, opened 2020-09-08
PIC-730 Back link not working

theDustRoom:
Fixed Back link from match defendant screen
Included BDD tests


katie-cawthorne (approved):
👍

PR #139, opened 2020-09-08
PIC-724 Possible nDelius record feature working incorrectly

theDustRoom:
Fixed display of possible matches on case summary
Updated mock data to be more inline with live data


theDustRoom:
Charan spotted I'd missed where this issue remained on case summary.

katie-cawthorne (approved):
👍

PR #142, opened 2020-09-09
Suppress recently added and removed tabs when no relevant cases exist

theDustRoom:
From a discussion with Charan earlier, we had discussed previously during testing that the "Recently added" and "Recently removed" tabs on the case list should be suppressed when there are no relevant cases.

This work is more urgent now as the matcher is no linger returning any "Recently removed" cases during the tactical release so I've completed this now to make sure the tab no longer shows, but will once we get to the strategic release and this data is then populated correctly.

### DIFF
--- a/server/views/case-list.njk
+++ b/server/views/case-list.njk
@@ -90,16 +90,39 @@
     } %}
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


katie-cawthorne (approved):
👍

PR #144, opened 2020-09-09
Pic 679 prototype alignment

katie-cawthorne:
2 small changes following meeting with Charan and Alastair regarding PIC-679 (prototype front end alignment)
- suppress `pac-key-details-bar__bottom-block` if there's no probation status
- On view `case-summary-record-order` moved back link out of main content and into backLink block so displayed the same as other pages


theDustRoom (approved):
👍

PR #145, opened 2020-09-09
:bug: PIC-726: fixed invalid date

katie-cawthorne:
Following investigations with Alastair it has been decided to use sentence startDate instead of convictionDate, due to convictionDate displaying null for cases over 5 years old.

- Replaced convictionDate with sentence.startDate
- Updated mock data to reflect live sentence data


theDustRoom (approved):
LGTM

PR #146, opened 2020-09-10
PIC-719 Current order start date

theDustRoom:
Corrected current order start date
Also corrected issue with order length calculation


katie-cawthorne (approved):
👍

PR #153, opened 2020-09-15
PIC-655 No cases match filters

theDustRoom:
Display no matching cases when filtered list is empty


katie-cawthorne (approved):
👍

PR #155, opened 2020-09-15
PIC - 704 : Returning order length in order title

sarahyjja:
Probably the testing/BDD had to be check as I wasn't sure about what to do in it...


public/stylesheets/app/_header.scss
```diff
@@ -9,3 +9,7 @@
   cursor: pointer;
   text-decoration: underline !important;
 }
+
+.heading-count-order{
```
theDustRoom, line 13: You should prefix `pac` to additionally required classnames. So this should be `pac-heading-count-order` though I'm not sure you need this additional class as you could probably use one of the GOV.UK caption classes:

See: https://design-system.service.gov.uk/styles/typography/

public/stylesheets/app/_header.scss
```diff
@@ -9,3 +9,7 @@
...
+  color: #505a5f;
```
theDustRoom, line 14: You should use the helper function to return the relevant colour, in this case this would be `govuk-color("dark-grey");`

See: https://design-system.service.gov.uk/styles/colour/

server/views/case-summary-record.njk
```diff
@@ -31,26 +31,35 @@
...
 
+            <div class="govuk-grid-row">
+                <div class="govuk-grid-column-full">
+                    <h1 class="govuk-heading-l">Probation record</h1>
+                </div>
+            </div>
+
             {% if activeOrders | length %}
-                <h2 class="govuk-heading-m">Current orders ({{ activeOrders | length }})</h2>
+                <h2 class="govuk-heading-m">Current orders <span class="govuk-caption-l govuk-!-font-weight-bold govuk-!-display-inline-block">({{ activeOrders | length }})</span></h2>
```
theDustRoom, line 41: I was able to switch out the additional class and include govuk frontend styles.

server/views/case-summary-record.njk
```diff
@@ -91,17 +100,7 @@
             {% endif %}
 
         </div>
-        <div class="govuk-grid-column-one-third">
```
theDustRoom, line 94: This needs to still be here but only display for current offenders.

theDustRoom (changes requested):

server/views/case-summary-record.njk
sarahyjja: cool

theDustRoom (approved):
👍

PR #158, opened 2020-09-17
Updated Node.js versions to security releases

theDustRoom:
Project now requires the use of Node.js 12.18.4 or 14.11.0 which include important security updates.


sarahyjja (commented):
coolio!

PR #165, opened 2020-09-22
Improved backward navigation support

theDustRoom:


theDustRoom:
I noticed that unlinking a defendant had an incorrect back/cancel URL.
Looking into it I decided to safely store the previous URL so that we can better support backward navigation.
It only stores the URL on GET and only updates if the URL is different (to catch page refreshes).

theDustRoom:
Tests are passing locally but failing on CI. I'll investigate...

PR #166, opened 2020-09-23
PIC-655 No Filtered Cases

theDustRoom:
Updated to display 0 cases when no filtered cases


sarahyjja (approved):
coolio

PR #180, opened 2020-09-29
:lock: Update IP whitelist to include HMCTS Magristrates' courts

markberridge:
Added the CIDR ranges which cover court users on DOM1 machines, accessing the service from an HMCTS Magistrates' Court network.


theDustRoom (approved):
👍

PR #184, opened 2020-09-30
:bug: PIC-361: Fixed missing pss requirements and licence conditions on probation record

katie-cawthorne:
- Added correct reference to pss requirements and licence conditions
- Updated mock data to align with court-case-service


sarahyjja (approved):
👍

PR #190, opened 2020-10-05
Pic798 - Possible nDelius flag not appearing on case list for some cases and missing banner at case level

sarahyjja:


theDustRoom (approved):
LGTM

PR #201, opened 2020-10-14
Pic 288 risk register

katie-cawthorne:
Mock data for registrations was structured under 'registrations' where as real data isn't so I've updated the mock data to the correct format.
Updated the endpoint and view to work with data structure


theDustRoom (approved):
👍

PR #202, opened 2020-10-14
Pic 288 risk register

katie-cawthorne:
- Small change...noticed in prod that when notes are empty the '-' wasn't being displayed, added code to fix this
- Updated registration mocks to add in empty notes field


theDustRoom (approved):
LGTM

PR #203, opened 2020-10-14
✏️   1 possible record banner content change

sarahyjja:
https://dsdmoj.atlassian.net/browse/PIC-808


theDustRoom (changes requested):
This introduces unnecessary duplication. You should change only the section necessary to achieve the corrected display in the UI.

theDustRoom (approved):
👍

PR #204, opened 2020-10-14
Pic 288 risk register

katie-cawthorne:
When notes are empty prod not displaying '-' when inactive registrations (did the fix in previous commit for active registrations but forgot about inactive 🤦🏻‍♀️)


theDustRoom (approved):
👍

PR #211, opened 2020-10-21
PIC-807 Ignore filters on added & removed case list views

theDustRoom:
:bug: Ensure filters are not applied to added and removed cases


sarahyjja (approved):
good

PR #212, opened 2020-10-21
PIC-878 Case List Navigation

theDustRoom:
:sparkles: Reconfigured case list navigation and initial support fo dynamic working days.


sarahyjja (approved):
I think I get it :) coolio

PR #213, opened 2020-10-26
✏️ Amend matcher call to action banner for single cases

sarahyjja:
https://dsdmoj.atlassian.net/browse/PIC-864

### DIFF
--- a/server/views/case-list.njk
+++ b/server/views/case-list.njk
@@ -157,10 +157,14 @@
 
     {% if params.unmatchedRecords > 0 %}
         {%- from "components/inset-text/macro.njk" import insetText -%}
+        {% set defendantCount = 'defendant' if params.unmatchedRecords <= 1 else 'defendants' %}
+        {% set recordCount = 'record' if params.unmatchedRecords <= 1 else 'records' %}
+        {% set matchCount = 'matches' if params.unmatchedRecords <= 1 else 'match' %}
+        {% set needCount = 'needs' if params.unmatchedRecords <= 1 else 'need' %}
         {{ insetText({
-            linkText: "Review defendants with possible nDelius records",
+            linkText: "Review " + defendantCount + " with possible nDelius " + recordCount,
             link: '/' + params.courtCode + "/match/bulk/" + params.date,
-            text: params.unmatchedRecords + (' defendants' if params.unmatchedRecords > 1 else ' defendant') + " partially matches existing records and needs review."
+            text: params.unmatchedRecords + " " + defendantCount + " partially " + matchCount + " existing records and " + needCount + " review."
         }) }}
     {% endif %}
 
--- a/server/routes/index.js
+++ b/server/routes/index.js
@@ -64,6 +64,7 @@ module.exports = function Index ({ authenticationMiddleware }) {
     }
     req.session.currentView = params.subsection
     req.session.caseListDate = params.date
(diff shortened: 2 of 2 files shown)


server/views/case-list.njk
```diff
@@ -157,10 +157,14 @@
 
     {% if params.unmatchedRecords > 0 %}
         {%- from "components/inset-text/macro.njk" import insetText -%}
+        {% set defendantCount = 'defendant' if params.unmatchedRecords <= 1 else 'defendants' %}
```
jhevans, line 160: `defendantPluralised` might be more descriptive than `Count` for this and the 3 variables below

integration-tests/integration/features/case-list.feature
```diff
@@ -14,7 +14,7 @@ Feature: Case list
     And I should see the caption with the relevant court
     And I should see sub navigation with default dates
     And I should see a timestamp of the most recent Libra data
-    And I should see that 2 defendants have possible nDelius records
```
jhevans, line 17: I think there should be an addition here, as it stands we're just not running this step by the looks of it, same goes for the other changes in this feature file.

integration-tests/integration/features/case-list/case-list.spec.js
```diff
@@ -21,10 +21,10 @@ And('I should see sub navigation with default dates', () => {
   cy.get('.pac-pagination-link--text').eq(0).contains('Next day')
 })
 
-And('I should see that {int} defendants have possible nDelius records', $int => {
+And('I should see that {int} defendant have possible nDelius records', ($int, $string) => {
```
jhevans, line 24: The `$string` parameter is never set in the step template (see suggestion for what I think it should be)- because of the other comment I left about the feature file I don't think this code is ever getting called which is why it hasn't caused any errors.

```suggestion
And('I should see that {int} {string} have possible nDelius records', ($int, $string) => {
```

integration-tests/integration/features/case-list/case-list.spec.js
```diff
@@ -21,10 +21,10 @@ And('I should see sub navigation with default dates', () => {
...
   cy.get('.pac-inset-text').within(() => {
-    cy.get('.govuk-link').contains('Review defendants with possible nDelius records').should('exist')
-    cy.get('.govuk-body').contains(`${$int} defendants partially matches existing records and needs review.`).should('exist')
+    cy.get('.govuk-link').contains(`Review ${$string} with possible nDelius ${$string}`).should('exist')
+    cy.get('.govuk-body').contains(`${$int} ${$string} partially ${$string} existing ${$string} and ${$string} review.`).should('exist')
```
jhevans, line 27: In this  case  I think multiple parameters are needed where we've got `$string` as there's different values. These will all need to be passed in from the template as well.

jhevans (changes requested):
There's some changes needed in the feature files as I don't think the steps are running as intended. Give me a shout if you need any help

server/views/case-list.njk
sarahyjja: 👍

integration-tests/integration/features/case-list/case-list.spec.js
sarahyjja: 👍

jhevans (approved):

PR #217, opened 2020-11-03
💄PIC-844 : Risk register date transform

sarahyjja:
The column width change seems to have been done previously...
https://dsdmoj.atlassian.net/browse/PIC-844


katie-cawthorne (approved):
Nice one 🙂

PR #219, opened 2020-11-03
:art: PIC-842: Updated to use sentence endDate field

katie-cawthorne:
- Updated to use conviction.sentence.endDate field instead of conviction.endDate field
- Updated mock data


sarahyjja (approved):
coolio :)

PR #222, opened 2020-11-05
PIC-785 Download Files Issue

theDustRoom:
:bug: Corrected issue where files would not download or would be corrupted


katie-cawthorne (approved):
Looks good

PR #226, opened 2020-11-12
PIC-881 View Case List on Sunday

theDustRoom:
:sparkles: Viewing case list on Sunday redirects to Monday

When viewing the case list on Sunday the application redirects to Monday.
In this case, the page title is changed from "Tomorrow" to "Monday DD MMMM".
Navigating back to case list from case summary functions correctly.
Unit tests include date mock to successfully test.

**N.B.** As this is a Node.js application we cannot mock the date for integration tests, therefore none are written and if the tests are run on a Sunday then we will have failing tests. There doesn't appear to be anything we can do about this.


katie-cawthorne (approved):
Nice one 👍

PR #236, opened 2020-11-26
:art: PIC-938: Updated OU code to 5 digits

katie-cawthorne:
- Removed last 2 digits from hardcoded OU code throughout app
- Updated integration tests
- Removed redundant environment variables from when we were testing with just one court
- Updated regex in mocks to accept 5 characters instead of 7


theDustRoom (approved):
BOOM

PR #239, opened 2020-12-02
:lipstick: PIC-953: Updated case list h1

katie-cawthorne:


sarahyjja (approved):
I think this ticket was inside mine too (pic-921)

PR #241, opened 2020-12-07
PIC-305 UI Components

theDustRoom:
Pulled out inline JavaScript for the filter component and included support for external JavaScript within the application.
Also added minify to process public JavaScript and CSS and included in build script.
Changed to use minified JavaScript and CSS.


sarahyjja (approved):
good good

PR #246, opened 2020-12-10
PC-929 Key Details Bar Refinement

theDustRoom:
Having spoken with Charan we've agreed (now that we will have PNC) to make the Probation status always appear on the line below the name.

I've used Flexbox as we only have to go back to IE11 and this seems best suited to achieve what we're after.


katie-cawthorne (approved):
LGTM 👍

PR #251, opened 2020-12-15
:construction_worker: Suppress snyk on build

theDustRoom:
I’ve removed the snyk image scan from the build process to allow the docker image to be pushed.

I’ve then added the image_scan to run in parallel with deploy_dev so that we can still be alerted.


katie-cawthorne (approved):
LGTM

PR #252, opened 2020-12-15
:lipstick: PIC-990: Added properCase filter

katie-cawthorne:
- Small change, noticed in sprint planning that the matched defendant names were displaying with incorrect capitalisation as missing `properCase` nunjucks filter, so added that.
- Also updated the mock data to reflect real data


theDustRoom (approved):
LGTM 👍

PR #254, opened 2020-12-16
:sparkles: PIC-866: Removed defendant title

katie-cawthorne:
- Added custom nunjucks filter to check name for titles using regex and replace with empty string if present
- Updated mock data to include titles in `defendantName` so reflects real data and can make sure the filter works correctly.

### DIFF
--- a/integration-tests/integration/features/matching.feature
+++ b/integration-tests/integration/features/matching.feature
@@ -357,7 +357,7 @@ Feature: Matching defendants to nDelius records
     When I click the "Unlink record from defendant" button
 
     Then I should be on the "Case summary" page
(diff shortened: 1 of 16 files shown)


sarahyjja (dismissed):
coolio!

theDustRoom (commented):
We're going to discuss this change tomorrow in standup as I notice that the ticket says that `defendantName` will be blank when we get to the strategic release and we need to handle this throughout the service.

theDustRoom (approved):
After the conversation with Chris we're confident that defendantName will still be populated in the strategic release.

We'll create a new ticket to make all uses of defendant name throughout the front end consistent.

PR #256, opened 2020-12-16
PIC-919 Suppress Filters

theDustRoom:
:children_crossing: Suppress filters when there are no cases


sarahyjja (approved):
coolio

PR #260, opened 2020-12-17
BUGFIX Issue with relative links and change to court name

theDustRoom:
:adhesive_bandage: Strip trailing slash to make sure all links can be relative (and preserve any querystring)
:speech_balloon: Fixed incorrect court name


sarahyjja (approved):
coolio

PR #261, opened 2020-12-18
💄 Fix no wrap Beta banner

sarahyjja:


theDustRoom (changes requested):
This ticket is related only to narrow widths so for standard desktop the header should be as it was previously with the positioning of the beta banner gracefully wrapping at lower resolutions/window width.

theDustRoom (approved):
LGTM

PR #275, opened 2021-01-04
PIC-1007 Day Navigation Issue

theDustRoom:
Day navigation was configured to use relative URLs but this causes an issue if you navigate to the **Recently added** or **Recently removed** tabs on the case list view.

I have implemented a _baseUrl_ value which is passed to the day navigation component in order to correct the URL.

:bug: Fixed issue with day navigation


sarahyjja (approved):
coolio!

PR #278, opened 2021-01-04
PIC-989 Case Summary Performance

theDustRoom:
:zap: Improved performance by removing unnecessary API calls


sarahyjja (approved):
oki doki

PR #285, opened 2021-01-13
PIC-1033: Added key details banner on match-defendant view when comin…

katie-cawthorne:
…g from case-summary

- Added key details banner on match-defendant view when coming from case-summary.
- Key details banner will not display on match-defendant view when coming from `/match/bulk` as per acceptance criteria.

### DIFF
--- a/integration-tests/integration/common/common.spec.js
+++ b/integration-tests/integration/common/common.spec.js
@@ -256,3 +256,11 @@ And('I should see the text input hint {string}', $string => {
 And('I enter {string} into text input with id {string}', ($string, $id) => {
   cy.get(`#${$id}`).type($string)
 })
+
+And('I should not see the key details banner', () => {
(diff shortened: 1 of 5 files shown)


sarahyjja (approved):
coolio :)

theDustRoom (changes requested):
We could do with some BDD around this but other than that it's noice 👍

theDustRoom:
I've just realised that this change is in the wrong file, it should be within match-manual.njk

katie-cawthorne:
Ok, I'll have a look

katie-cawthorne:
I've removed the key details banner from `match-defendant.njk` and added to `match-manual.njk` (sorry got confused which page it was)
Updated the BDD tests

theDustRoom (approved):
LGTM 👍

PR #288, opened 2021-01-18
Pic 1033 key details banner

katie-cawthorne:
Alastair recognised that the `Link nDelius record` button on the key details banner was displaying on match-manual view. This wasn't mentioned on the ticket but the button is not required on this page, so I have removed the button as requested.


sarahyjja (approved):
looks good to me :)

PR #301, opened 2021-01-29
Unsupported court code causes front end to crash

theDustRoom:
As the court codes and names are currently hardcoded, changing this code in the URL causes  a crash in the frontend.

This fixes the issue but in these instances, the court name will of course not be displayed.

:bug: Unsupported court code causes front end to crash


sarahyjja (approved):
👍

PR #311, opened 2021-02-23
PIC-1098 Case List Snapshot

theDustRoom:
:sparkles: Included latest snapshot in case list request


katie-cawthorne (approved):
Nice one 👍

PR #318, opened 2021-03-05
Content change: Bulk records list screen when 1 defendant on list

sarahyjja:


server/views/match-records.njk
```diff
@@ -37,11 +37,11 @@
         matchName: session.confirmedMatch.name,
         matchType: session.confirmedMatch.matchType
     }) }}
-
     <h1 class="govuk-heading-l">Defendants with possible NDelius records</h1>
 
     {% if tableData.rows | length %}
-        <p class="govuk-heading-m">{{ tableData.rows | length }} defendants partially match with existing records.</p>
+        {% set oneDefendant = 'defendant' if tableData.rows === 1 else 'defendants' %}
+        <p class="govuk-heading-m">{{ tableData.rows | length }} {{ oneDefendant }} partially match with existing records.</p>
```
theDustRoom, line 44: I would change this to `{{ 'defendant' if tableData.rows === 1 else 'defendants' }}` instead of setting as a value above as it's not used anywhere else.

theDustRoom (changes requested):

server/views/match-records.njk
theDustRoom: Actually, I don't think this will work as it's not checking the length of `tableData.rows` so this will never be 1

theDustRoom (approved):
LGTM

PR #321, opened 2021-03-08
PIC-1166 Additional courts

theDustRoom:
:lipstick: Split select courts UI into two columns
:wrench: Added short names for courts in config


sarahyjja (approved):
coolio!

PR #324, opened 2021-03-11
:adhesive_bandage: PIC-1173: added cookie-session

katie-cawthorne:
Replaced express-session with cookie-session due to issues with express-session memory store capability of keeping the session info in memory in one of the pods. If another request comes in and hits another pod then it won't find the session and will consider the user logged out / or have to have an invalid session. This is causing issues for us when we increase the amount of pods.

### DIFF
--- a/app.js
+++ b/app.js
@@ -4,11 +4,9 @@ const config = require('./config')
 const express = require('express')
 const compression = require('compression')
 const cookieParser = require('cookie-parser')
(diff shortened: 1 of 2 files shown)


theDustRoom (approved):
LGTM

theDustRoom:
After merging this we found issues with auth so I’ve reverted the changes in the builds.

PR #326, opened 2021-03-15
Pic 1173 add cookie session

katie-cawthorne:
- I've amended the cookie-session config to increase session expiry to 2 hours.
- Removed the line of code `req.session.destroy()` as I think this may have been causing us the issue with auth.


sarahyjja (approved):
👍

PR #329, opened 2021-03-16
PIC-1096 Probation Status

theDustRoom:
As per discussion with Chris, we need to change the probation status to null when unlinking a defendant

:bug: Fixed probation status when unmatching defendant


katie-cawthorne (approved):
LGTM

PR #330, opened 2021-03-16
Pic 1181 - Getting "Undefined" when no forename is in data

sarahyjja:


mappings/case-list-full.json
```diff
@@ -281,6 +279,11 @@
           ],
           "defendantSex": "M",
           "defendantName": "MR Webb Mitchell",
+          "name": {
+            "title": "Mister",
+           "forename1": "webster",
```
katie-cawthorne, line 284: We could probably do with an example in the mock data where there is no `forename1` or `surname` just to show that it defaults to `defendantName`.

sarahyjja: Modification has been done to `MISS FELICIA Villarreality`

sarahyjja (commented):
Change made on `MISS FELICIA Villarreality`

katie-cawthorne (approved):
LGTM ✨

PR #332, opened 2021-03-16
PIC-1173: Removed cookie-session

katie-cawthorne:
I've raised a few PR's when adding the cookie-session so thought it would be easier to remove cookie-session in a new PR rather than reverting the others.


theDustRoom (approved):
👍🏻

PR #339, opened 2021-03-22
PIC-1196: added reference to ingress certs

katie-cawthorne:
I've added the ingress certificates in court-probation-dev and court-probation-preprod, so this PR references them in the app. We already have a certificate in prod. We need this for redis to work correctly.


sarahyjja (approved):
coolio

PR #345, opened 2021-03-24
Ga security policy

sarahyjja:


theDustRoom:
You need to pass the nonce to the template when the route is not using the defaults middleware, `/select-court` and `/user-guide` are the only routes affected but they still shouldn't use defaults.

theDustRoom (changes requested):
Changes needed to `/user-guide` route.

theDustRoom (approved):
LGTM

PR #349, opened 2021-03-25
:wrench: added USER_PREFERENCE_SERVICE_URL env var

katie-cawthorne:
Build failed as missing USER_PREFERENCE_SERVICE_URL env var in `.envs.tpl`


theDustRoom (approved):
Yep, knew I'd missed something - thanks @katie-cawthorne

PR #353, opened 2021-04-01
Store secrets for different environments

sarahyjja:


server/views/partials/layout.njk
```diff
@@ -37,7 +37,7 @@
             j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
             'https://www.googletagmanager.com/gtm.js?id='+i+dl;var n=d.querySelector('[nonce]');
             n&&j.setAttribute('nonce',n.nonce||n.getAttribute('nonce'));f.parentNode.insertBefore(j,f);
-          })(window,document,'script','dataLayer','UA-106741063-19');</script>
+          })(window,document,'script','dataLayer',{{ googleAnalyticsKey }});</script>
```
katie-cawthorne, line 40: You'll need to enclose the googleAnalyticsKey in single speech marks like '{{ googleAnalyticsKey }}', I know that looks weird but don't think it will work without it.

katie-cawthorne (changes requested):
Excellent other than just putting the variable in single speech marks 👍

server/views/partials/layout.njk
sarahyjja: awes! :) will do!

katie-cawthorne (approved):
Fab ✨

katie-cawthorne (approved):
👍

PR #357, opened 2021-04-08
Node security release

theDustRoom:
Update requirement for Node.js April 2021 Security Release.
Updated to use new next-gen CircleCi convenience Docker images.

See https://nodejs.org/en/blog/vulnerability/april-2021-security-releases/


sarahyjja (approved):
👍

PR #359, opened 2021-04-12
Rename court cookie

theDustRoom:
As we're putting the user preferences service live tonight I realised that if the user currently has the court cookie then they won't get the new onboarding journey and will therefore not store their chosen court(s) in the user preferences service.

In order to ensure all users will get the new experience I have chosen to rename the cookie to currentCourt.

I have also included some code to remove the old court cookie to make sure it's not hanging about on all our current users' machines.

We can remove this code at a later date.

♻️ Renamed the court cookie
🚧 Included code to remove the old cookie which will need to be removed at a later date.

### DIFF
--- a/server/routes/index.js
+++ b/server/routes/index.js
@@ -33,7 +33,11 @@ module.exports = function Index ({ authenticationMiddleware }) {
 
   router.get('/', (req, res) => {
     const { cookies } = req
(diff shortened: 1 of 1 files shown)


katie-cawthorne (approved):
👍

PR #360, opened 2021-04-13
Add phase 2 courts to user preferences type ahead

sarahyjja:
Some of the courts were already existing in the list. Added the missing ones.


katie-cawthorne (approved):
👍

PR #361, opened 2021-04-14
:zap: PIC-1175: Added new single conviction endpoint

katie-cawthorne:
- added functionality to point to single conviction endpoint from court-case-service, so we can use this instead of the costly probation-record endpoint when requesting data for a single conviction
- amended current single conviction endpoint to use the new single conviction endpoint 
- added unit tests for new functions
- added pact test for getConviction
- added BDD data for single conviction


theDustRoom (approved):
Nice work Katie 👍

sarahyjja:
Yes I want to say the same! 🤩

PR #368, opened 2021-04-19
♿️ A11y fixes

theDustRoom:
After the a11y review we needed to address a few minor issues.
The Case list tabs do not function as tabs so this has been changed to use standard navigation elements with additional aria-label text to make this clearer.
Page titles have been updated accordingly on the case list page(s).
Defendant name links have additional aria-label text to make this a little clearer to screen readers.
Removed erroneous } in _header.html_ which caused an a11y issue not detected by our integration tests.


katie-cawthorne (approved):
👍

PR #374, opened 2021-04-26
:lipstick: PIC-1287: added brackets around url

katie-cawthorne:
- added brackets around url as requested by Alastair


sarahyjja (approved):
👍

PR #376, opened 2021-04-26
Cookie fix

theDustRoom:
:bug: Fix cookie expiry

The expiry for cookies stored for current court and analytics preference will now extend from the last visit to the application.
The user would now have to not use the application for 6 months to lose the cookie.


sarahyjja (approved):
👍

PR #378, opened 2021-04-27
:loud_sound: PIC-1324: Added start up log message

katie-cawthorne:
Applicationinsights automatically collects bunyan logs, so the log message will display in appinsights and this should give us visibility of our deployment frequency.


theDustRoom (approved):
👍

PR #380, opened 2021-04-27
PIC-1339 Fox Google Analytics 

theDustRoom:
After testing on Dev CSP was still blocking GA. I've updated this based on a blog post I found, looks like a belt & braces approach but...

:bug: Updated CSP to allow all GA connections/scripts


sarahyjja (approved):
🦊 👍

PR #393, opened 2021-05-11
Pic 1372 add pre sentence record to filter

katie-cawthorne:
- Added new probation status `Pre-sentence record` to filter. They are dynamic so the only options shown in the filters are what we have in the data, so this change will be ok to add before the back end work has been done - if we have no Pre-sentence record coming back in the data then it won't be an option in the filter.
- Added a BDD test to demonstrate that it works
- Updated mock data to include an example of `Pre-sentence record`

### DIFF
--- a/server/utils/getCaseListFilters.js
+++ b/server/utils/getCaseListFilters.js
@@ -1,7 +1,7 @@
 module.exports = (caseListData, selectedFilters) => {
   const availableProbationStatuses = [...new Set(caseListData.map(item => item.probationStatus))]
   const probationStatuses = []
-  const statusOrder = ['Current', 'Previously known', 'No record']
+  const statusOrder = ['Current', 'Previously known', 'No record', 'Pre-sentence record']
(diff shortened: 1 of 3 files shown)


server/utils/getCaseListFilters.js
```diff
@@ -1,7 +1,7 @@
 module.exports = (caseListData, selectedFilters) => {
   const availableProbationStatuses = [...new Set(caseListData.map(item => item.probationStatus))]
   const probationStatuses = []
-  const statusOrder = ['Current', 'Previously known', 'No record']
+  const statusOrder = ['Current', 'Previously known', 'No record', 'Pre-sentence record']
```
theDustRoom, line 4: The order should be `['Current', 'Pre-sentence record', 'Previously known', 'No record']`

theDustRoom (changes requested):
The ticket shows the order of the filters differently, I've checked with Alastair and it should be in that order.

PR #401, opened 2021-05-13
PIC-1391 GA page load times

theDustRoom:
The default load times uses 1% of the visitors for the measurement by default so the average is 0.

Looking at https://developers.google.com/analytics/devguides/collection/gtagjs/user-timings this looks to do what we need, so this is what the PR is for.

I've included it at the page end to make sure the event fires after the whole page has loaded.


sarahyjja (approved):
coolio

PR #403, opened 2021-05-19
PIC-1400 pact alignment

theDustRoom:
Chris has confirmed that all PACT tests are passing on the broker now.


katie-cawthorne (approved):
LGTM

PR #405, opened 2021-05-24
PIC-1405 Cache calls to user/me and remove redundant user details call

theDustRoom:


server/data/authClientBuilder.js
```diff
@@ -3,6 +3,13 @@ const Agent = require('agentkeepalive')
 const { HttpsAgent } = require('agentkeepalive')
 const logger = require('../../log')
 const config = require('../../config')
+const csRedis = require('cache-service-redis')
+const redisCache = new csRedis({
+  port: config.redis.port,
+  hostname: config.redis.port,
+  auth: config.redis.password
```
theDustRoom, line 10: When testing, if you add `verbose: true` here you will see the cache output in your console.

server/services/userService.js
```diff
@@ -7,17 +7,10 @@ module.exports = function createUserService (authClientBuilder) {
 
       // Get the user record from Auth to determine whether this is an auth-only or Nomis account
       const user = await authClient.getUser()
-      let nameForDisplay = null
```
theDustRoom, line 10: Sorted out some pretty awful code here

server/services/userService.js
```diff
@@ -47,26 +40,8 @@ module.exports = function createUserService (authClientBuilder) {
     }
   }
 
-  async function getUserDetails (context) {
```
theDustRoom, line 50: We don't use this and it was being called on each route.

package.json
```diff
@@ -78,6 +78,7 @@
     "axios-middleware": "0.3.1",
     "bunyan": "1.8.15",
     "bunyan-format": "0.2.1",
+    "cache-service-redis": "2.0.2",
```
theDustRoom, line 81: As we have Redis I figured it's better to use this than an in-memory cache

sarahyjja (approved):

PR #410, opened 2021-05-26
PIC-1431 Support awaiting PSR in consumer mocks and PACT tests

theDustRoom:
👽 Updated to support `awaitingPsr` flag.
👽 Improved PUT case schema.


katie-cawthorne (approved):
LGTM

PR #425, opened 2021-06-10
Add the CRN to matched cases on case list views

sarahyjja:


theDustRoom (changes requested):
You should rollup the HTML into the crnDisplay value so that when there's no CRN we don't have any empty/redundant HTML elements in the DOM.

theDustRoom (approved):
👍

PR #426, opened 2021-06-10
Pic 1159: added in custody status 

katie-cawthorne:
✨  Added in custody status to probation-record and order page
🤡  Updated mocks
✅  Updated pact test and schema to add custodialType to probation-record call


sarahyjja (approved):
coolio!

PR #433, opened 2021-06-17
:white_check_mark: PIC-1439: removed pssRequirements data

katie-cawthorne:
Removed the pssRequirements data to fix PACT, this isn't expected to return data in court-case-service when returning requirements and licenceConditions


chris-faulkner-mojdt (approved):
LGTM

PR #438, opened 2021-06-21
End-of-line normalization

theDustRoom:
This should enforce text end-of-line on all relevant files.

🔧 Introduce end-of-line normalization


sarahyjja (approved):
👍

PR #440, opened 2021-06-22
Browser cache static assets

theDustRoom:
Reconfigured to cache static assets for 30 days (lighthouse recommendation), JavaScript cache set to 24 hours (seems reasonable).
CSS and app specific JavaScript uses the APP_VERSION environment variable to set a version number, this ensures any changes will not be cached by the browser. As these are versioned, they use the asset cache so if no changes are made, they will be cached for 30 days.

N.B. Locally we will be caching assets so you will have to clear cache and hard reload if you make CSS or client-side JavaScript changes.


katie-cawthorne (approved):
LGTM

PR #445, opened 2021-06-24
Pic-1510: Uppercase/lowercase formatting on matcher journey

sarahyjja:


package.json
```diff
@@ -108,7 +108,7 @@
     "axe-core": "4.2.3",
     "cypress": "7.6.0",
     "cypress-axe": "0.12.2",
-    "cypress-cucumber-preprocessor": "4.1.2",
+    "cypress-cucumber-preprocessor": "^4.0.0",
```
theDustRoom, line 111: Don't think you meant to do this?

theDustRoom (changes requested):
This will need to be included everywhere we display the defendant name to ensure we catch them all. Not sure if that's covered by this ticket.

package.json
sarahyjja: I think it came when I was doing all these weird manipulation with npm :/

theDustRoom (approved):
LGTM

PR #456, opened 2021-07-15
PIC-1535 - Auth error route

theDustRoom:
On the request for access email, I noticed that the attached screenshot had no styling applied, this is because the `nonce` and `appVersion` are stored in res.locals incorrectly. This PR fixes this.

🐛 Fix error loading assets on authError


sarahyjja (approved):
cool

PR #460, opened 2021-07-22
PIC-1503 Cascade failure

theDustRoom:
Issue is that Axios does not have a default timeout 😒

This change specifies configurable timeouts for server requests so that the application will not stall and subsequently fail if the backend becomes unresponsive.

🥅 Set and handle request timeouts
🎨 Minor refactors
🔧 Included timeouts in config


sarahyjja (approved):
okidok

PR #468, opened 2021-08-09
🔥 Unhide author on probation record page

jhevans:
The author is now being populated correctly so this will show it again


sarahyjja (approved):
coolio

PR #471, opened 2021-08-12
:bug: PIC-1579: fix post release dates bug

katie-cawthorne:
I also noticed a bug when the offender is in custody so have fixed that also.

Signed-off-by: katie-cawthorne <katie.cawthorne@digital.justice.gov.uk>

### DIFF
--- a/server/views/case-summary-record-order.njk
+++ b/server/views/case-summary-record-order.njk
@@ -25,6 +25,7 @@
 
 {% if communityData.sentenceDetails.custody.custodialType.code %}
     {%- set custodialOrder = communityData.sentenceDetails.custody -%}
+    {%- set inCustody = communityData.sentenceDetails.custody.custodialType.description == 'In Custody'-%}
     {%- set isOnLicence = moment(custodialOrder.licenceExpiryDate, 'YYYY-MM-DD').isAfter() -%}
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


server/views/case-summary-record-order.njk
```diff
@@ -134,9 +135,9 @@
 
             <div class="govuk-grid-row">
                 <div class="govuk-grid-column-one-third">
-                    <p class="govuk-hint govuk-!-margin-bottom-0 qa-start-title">{{ "Post-release status" if custodialOrder else "Order started" }}</p>
+                    <p class="govuk-hint govuk-!-margin-bottom-0 qa-start-title">{{ "Post-release status" if custodialOrder and not inCustody else "Order started" }}</p>
                     <p class="govuk-body-l govuk-!-font-weight-bold govuk-!-margin-0 qa-start-date">
-                        {%- if custodialOrder -%}
+                        {%- if custodialOrder and not inCustody -%}
```
jhevans, line 140: `inCustody` is only ever used in this combination  
```if custodialOrder and not inCustody```

`custodialOrder and not inCustody` seems like a good candidate for  variable  itself, something like `hasLeftCustody`

jhevans (approved):
One suggestion to improve readability but happy to approve

server/views/case-summary-record-order.njk
katie-cawthorne: I was going to use the `inCustody` variable further down the page where we use the expression `if custodialType === "In Custody"`,  but forgot to update that but good point.

PR #486, opened 2021-09-01
PIC-1625 licence pss date

theDustRoom:
After discussion with Alastair we've determined that the data from the new endpoint (displayed in the table below) is correct so we have included this information in the top section and set it to not display if this is absent from the retuned data just incase.


sarahyjja (approved):
👍

PR #489, opened 2021-09-14
Pic 1578 update user guide

katie-cawthorne:
💄 Updated user guide page to remove sidebar navigation and follow gov.uk 'guide' format
💄 Added dash list style
✨ Embedded training videos - when running locally the videos won't display, YouTube has a domain allow list so we must access the embedded YouTube videos via a domain name, not locally...the videos display fine when running integration tests
💬 Updated content
✅ Added tests


theDustRoom (approved):
LGTM

PR #490, opened 2021-09-14
Fix PACT tests

theDustRoom:
Looks like the minor changes to Axios has caused this issue?

✅ Fix failing PACT tests


katie-cawthorne (approved):
LGTM

PR #492, opened 2021-09-21
PIC-1666 Display source of case to user

theDustRoom:
✨ Display source of case to user


katie-cawthorne (approved):
LGTM

PR #542, opened 2022-03-02
PIC-1922: CP display for case with same listing number

sarahyjja:


server/views/case-list.njk
```diff
@@ -42,11 +42,19 @@
...
     {% set listing = [] %}
+    {% set updatedListing = [] %}
     {% if item.source === "COMMON_PLATFORM" %}
         {{ listing.push('<ol class="govuk-list govuk-!-margin-bottom-0">') }}
-        {% for listNOfromCP in item.offences %}
-            {{ listing.push('<li>' + listNOfromCP.listNo | ordinalNumber | escape + '</li>' ) }}
+        {% for cpListNo in item.offences %}
+            {{ updatedListing.push(cpListNo.listNo) }}
         {% endfor %}
+        {% if updatedListing | first !== updatedListing | last %}
```
theDustRoom, line 51: Unfortunately, this doesn't work as expected. It only checks against the first and last list number so if you change any that aren't first or last to say, 2 then it still only displays "1st".

Need to ensure that these scenarios are covered with an integration test as well.

theDustRoom (changes requested):

theDustRoom (approved):

PR #546, opened 2022-03-08
Filter/Show contact details for CP cases

sarahyjja:


sarahyjja:
This PR solves also the ticket about "nationality" : https://dsdmoj.atlassian.net/browse/PIC-1647

integration-tests/integration/features/case-summary.feature
```diff
@@ -18,6 +18,45 @@ Feature: Case summary
...
+    And I should see the following summary list
+      | Name          | Mann Carroll                                                          |
+      | Gender        | Female                                                                |
+      | Date of birth | Unknown                                                               |
+      | Phone number  | 07 123 123 123 (Mobile) 020 456 456 456 (Home) 020 789 789 789 (Work) |
+      | Address       | 91 Barbey street Sheffield AD21 5DR                                   |
+
+    And There should be no a11y violations
+
     And There should be no a11y violations
```
jhevans, line 60: Looks like this line has been duplicated

schemas/get-case.schema.json
```diff
@@ -368,19 +367,48 @@
...
+      "required": [],
+      "properties": {
+        "mobile": {
+          "$id": "#/properties/defendantPhoneNumber/properties/mobile",
+          "type": [
+            "string",
+            "null"
+          ],
+          "title": "The mobile phoneNumber schema",
+          "description": "An explanation about the purpose of this instance.",
```
jhevans, line 388: Can we omit these descriptions if we have no more valuable information to add? Keeping this description boilerplate adds nothing so if it doesn't break anything we should get rid of it and keep the schema a bit cleaner.

jhevans (approved):
Couple of minor suggestions but mostly looks good 👍

PR #564, opened 2022-04-27
PIC-2118: Update and Delete offenders using offender endpoint

jhevans:


jhevans:
This is dev complete but local testing has shown up an unintended side effect, see [here](https://dsdmoj.atlassian.net/browse/PIC-2123) for more details. I'll hold off merging this until we've decided what to do about this.

rschallapalli (approved):

PR #566, opened 2022-05-09
Hotfix crash issue

theDustRoom:
In production, the pods are restarting due to the service crashing.

After checking the logs in k8s I found that when the health check fails, the service tries to render a 500 page but crashes because `.end()` is included. Removing this should correct the issue.


sarahyjja (approved):
coolio

PR #567, opened 2022-05-10
PIC-2141: ♿️ Accessibility: Incorrect  use of ARIA on the date picker

sarahyjja:


server/routes/middleware/healthcheck.js
```diff
@@ -3,7 +3,7 @@ const { apis: { courtCaseService }, settings: { healthTimeout } } = require('../
 
 const health = async (req, res, next) => {
   function unhealthy () {
-    res.render('error', { status: 500 })
+    res.render('error', { status: 500 }).end()
```
theDustRoom, line 6: Regression here: `.end()` was removed in order to correct a crash issue.

theDustRoom (changes requested):

PR #569, opened 2022-05-16
Court room numbers are appearing twice in the pick list fields

sarahyjja:


server/utils/nunjucksSetup.js
```diff
@@ -67,6 +67,13 @@ module.exports = (app, path) => {
     return sourceString.includes('Courtroom') ? sourceString.replace(/([A-Za-z 0]*)?/, '') : sourceString.replace(/([0]*)?/, '')
   })
 
+  // env.addFilter('courtRoomDisplay', (sourceString) => {
```
theDustRoom, line 70: We shouldn't be checking in commented code.

theDustRoom (changes requested):

PR #576, opened 2022-05-24
PIC-2124 fix duplicate court room numbers in court room filter

rschallapalli:


theDustRoom:
@rschallapalli This change should be implemented within `court-case-service` as anything downstream of this should not have to correct the duplicate courtroom identifiers. 

I discussed this with @markberridge and he agreed that this change should be made in the backend.

rschallapalli:
Fixing this on the backend 
"@rschallapalli This change should be implemented within court-case-service as anything downstream of this should not have to correct the duplicate courtroom identifiers.

I discussed this with @markberridge and he agreed that this change should be made in the backend."

PR #580, opened 2022-05-25
Included husky to lint and unit test before committing 

theDustRoom:
➕ Add Husky as dev depencency
🔨 Configured Husky


sarahyjja (approved):
oki

PR #589, opened 2022-06-01
PIC-2178 fix display of unmatched aliases

rschallapalli:


jhevans (dismissed):
Nice neat way of doing it 👍

jhevans:
@rschallapalli there's a linting issue and a failing integration test that need fixing but otherwise looks ok

server/views/match-defendant.njk
```diff
@@ -143,14 +143,8 @@
...
-
-                              {% for i in range(0, aliasDisplayLimit) -%}
-                                {% set aliasName = aliases[i].firstName + ' ' + (aliases[i].middleNames | join(' ') + ' ' if aliases[i].middleNames) + aliases[i].surname %}
-                                {% set markedAlias =  (aliasName | apostropheInName | properCase | removeTitle | markMatches(defendantFullName) | safe | emptyIfUnmarked )%}
-                                {{ ', ' if (not loop.first and markedAlias) }} {{ markedAlias }}
-                              {%- endfor %}
-
-                              {{ ['+ ', aliasesLength - 3, ' other name', ('s' if aliasesLength > 4), ' matched' ] | join('') if aliasesLength > 3  }}
+                              {% for i in range(0, aliasDisplayLimit) %}{% set aliasName = aliases[i].firstName + ' ' + (aliases[i].middleNames | join(' ') + ' ' if aliases[i].middleNames) + aliases[i].surname %}{% set markedAlias =aliasName | apostropheInName | properCase | removeTitle | markMatches(defendantFullName) | safe | emptyIfUnmarked %}{{ ', ' if (not loop.first and markedAlias)}} {{ markedAlias }} {%- endfor -%}
```
jhevans, line 146: That's a lot of stuff on one line, could we split it out or was that what was causing the problem?

jhevans (dismissed):

server/views/match-defendant.njk
rschallapalli: I had to do this to fix the failing integration test. In the for loop anything that's output is having leading white space and so the commas to are having that leading whitespace.  Spent more than half a day to figure this out.

Tried something else now to trim the leading spaces, by adding hyphens to the template snippets, which I also tried yesterday but at different places...  and it worked today 🤷‍♂️

jhevans (dismissed):

jhevans (approved):

PR #601, opened 2022-06-15
🐛 PIC-2042: Fix ordinal number filter for invalid integers

jhevans:


jhevans:
Good call to keep it consistent, will do 👍 The reason for creating a new file was simply for testing purposes. As `nunjucksSetup.js` only exports a single function there was no way of unit testing the individual filters without either refactoring to export an object from `nunjucksSetup.js` or moving the functions to be tested to a separate module, so I did the latter.

sarahyjja (approved):
👌

PR #602, opened 2022-06-20
♿️ PIC-2143: Accessibility: Visual headings not marked up (Filters)

sarahyjja:
Replace `<p>Filters / Applied filters</p>` by `<h2>Filters / Applied filters</h2>`

### DIFF
--- a/server/views/components/filter/template.njk
+++ b/server/views/components/filter/template.njk
@@ -3,7 +3,7 @@
 <section class="pac-filter-card govuk-!-margin-bottom-6">
     <form name="pac-filter-form" id="pac-filter-form" method="post">
 
-        <p class="govuk-heading-s">Filters</p>
+        <h2 class="govuk-heading-s">Filters</h2>
 
         {% for filter in params.filters %}
 
@@ -63,9 +63,9 @@
         {% endfor %}
 
         <div class="pac-selected-filters{% if not hasSelected %} moj-js-hidden{% endif %}">
-            <p class="govuk-body govuk-!-margin-bottom-0">
+            <h2 class="govuk-body govuk-!-margin-bottom-0">
                 <span class="govuk-!-font-weight-bold">Applied filters</span>
-                <a href="#" id="reset-filters" class="govuk-link govuk-link--no-visited-state govuk-body-s govuk-!-margin-left-2">Clear all</a></p>
+                <a href="#" id="reset-filters" class="govuk-link govuk-link--no-visited-state govuk-body-s govuk-!-margin-left-2">Clear all</a></h2>
 
             {% for filter in params.filters %}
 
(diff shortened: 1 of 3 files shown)


server/views/components/filter/template.njk
```diff
@@ -63,9 +63,9 @@
         {% endfor %}
 
         <div class="pac-selected-filters{% if not hasSelected %} moj-js-hidden{% endif %}">
-            <p class="govuk-body govuk-!-margin-bottom-0">
+            <h2 class="govuk-body govuk-!-margin-bottom-0">
                 <span class="govuk-!-font-weight-bold">Applied filters</span>
-                <a href="#" id="reset-filters" class="govuk-link govuk-link--no-visited-state govuk-body-s govuk-!-margin-left-2">Clear all</a></p>
+                <a href="#" id="reset-filters" class="govuk-link govuk-link--no-visited-state govuk-body-s govuk-!-margin-left-2">Clear all</a></h2>
```
jhevans, line 68: Should the header be applied to the 'Clear all' button as well here? I'd say probably not if we want to keep it consistent with the visual experience as it's smaller text.

jhevans: Or it might be an h3?

sarahyjja: Good thought @jhevans  :) I thought the same at the start but the Jira ticket mentioned explicitly to define it as an `h2` ...

Concerning the `Clear all` button, this one I believe should stay how it is as there is no comment related to it...

jhevans: Sorry that comment was badly written, let me try again 😆.

The `h2` has been applied around both the `Applied filters` text and the `Clear all` button -  I think it should only be around the `Applied filters` i.e.

```
<p class="govuk-body govuk-!-margin-bottom-0">
  <h2><span class="govuk-!-font-weight-bold">Applied filters</span></h2>
  <a href="#" id="reset-filters" class="govuk-link govuk-link--no-visited-state govuk-body-s govuk-!-margin-left-2">Clear all</a>
</p>
```

sarahyjja: Ah no worries ^^! I get what you mean though! I guess here, we cannot really do your proposal for accessibility reasons... As we trying to keep a hierarchy inside of the tags, imbricating an `<h2>` inside of a `<p>` might be unproductive as I mean not help the readability... I understand though that you would like to define the span only as an `<h2>` rather than the whole capsule, but we actually can't unwrap those as they are depending on each other. If it can help, I'm happy to jump on a call and demo it for you?

jhevans: Update: @sarahyjja attempted my suggestion but getting the markup right whilst keeping the visual layout the same hasn't been possible. Will approve. 👍

jhevans (approved):

PR #603, opened 2022-06-21
PIC-2175 FIx Multiple draft pre-sentence reports on the Probation Rec…

rschallapalli:
…ords tab

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,9 @@
     "ext": "js,html,njk,scss"
   },
   "standard": {
+    "env":  [
+      "jest"
+    ],
     "ignore": [
       "/public/**/*.js",
       "/integration-tests/config",
--- a/package.json
+++ b/package.json
@@ -46,9 +46,6 @@
     "ext": "js,html,njk,scss"
   },
   "standard": {
-    "env":  [
-      "jest"
-    ],
     "ignore": [
       "/public/**/*.js",
       "/integration-tests/config",
--- a/server/routes/index.js
+++ b/server/routes/index.js
@@ -274,10 +274,22 @@ module.exports = function Index ({ authenticationMiddleware }) {
 
     const crn = templateValues.data.crn
     const communityResponse = await getProbationRecord(crn, true)
+
+    const psrRequestedConvictions = communityResponse.convictions?.filter(conviction => conviction.active && conviction.awaitingPsr && conviction.psrReports && conviction.psrReports.length)
+      .map(conviction => {
+        const mainOffence = conviction.offences.find(offence => offence.main)
+        return {
+          psrReport: conviction.psrReports[0],
+          offence: mainOffence
+        }
+      })
+
     templateValues.params.showAllPreviousOrders = session.showAllPreviousOrders
     templateValues.data.communityData = {
       ...communityResponse
     }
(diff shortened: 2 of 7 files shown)


server/routes/index.js
```diff
@@ -274,10 +274,22 @@ module.exports = function Index ({ authenticationMiddleware }) {
...
+          psrReport: conviction.psrReports[0],
+          offence: mainOffence
+        }
+      })
+
     templateValues.params.showAllPreviousOrders = session.showAllPreviousOrders
     templateValues.data.communityData = {
       ...communityResponse
     }
+    templateValues.data.psrRequestedConvictions = psrRequestedConvictions
```
jhevans, line 291: Good call to put this in JS, the less logic we have in Nunjucks the better IMO.

jhevans (changes requested):

package.json
```diff
@@ -46,6 +46,9 @@
     "ext": "js,html,njk,scss"
   },
   "standard": {
+    "env":  [
```
rschallapalli, line 49: to fix this 
image

server/routes/index.js
```diff
@@ -274,10 +274,22 @@ module.exports = function Index ({ authenticationMiddleware }) {
...
+
+    const psrRequestedConvictions = communityResponse.convictions?.filter(conviction => conviction.active && conviction.awaitingPsr && conviction.psrReports && conviction.psrReports.length)
+      .map(conviction => {
+        const mainOffence = conviction.offences.find(offence => offence.main)
+        return {
+          psrReport: conviction.psrReports[0],
+          offence: mainOffence
+        }
+      })
+
```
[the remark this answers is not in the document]
rschallapalli: Added

package.json
jhevans: I've not seen this before, does this achieve the same thing as declaring globals [like so](https://github.com/ministryofjustice/prepare-a-case/blob/53c49d73351fbd6b0d32c61dc1ef5e01dc15de16/tests/services/community-service.test.js#L1) but globally?

If so is there a reason why we've not done this previously? Unlikely to happen but will this cause standard to accept `describe/it/etc` in production code, i.e. not in tests? @theDustRoom @sarahyjja any thoughts?

tests/routes/helpers.test.js
```diff
@@ -0,0 +1,31 @@
+const { getPsrRequestedConvictions } = require('../../server/routes/helpers')
+
+const testConvictions = require('../../mappings/community/convictions/D541487-convictions.json').response.jsonBody.convictions
```
jhevans, line 3: There's a risk that if this file changes then your unit tests will start failing, despite the fact the code is still working as intended. Would it be better to create the testConvictions locally to fully isolate the unit test?

package.json
theDustRoom: Yes, as the `describe/it/etc` are only ever in use within our tests we declare them at the start of the tests so that standard is happy and does not include these globals elsewhere when they're not needed.

jhevans: Cool, that makes sense. Probably best to go with something like `/* global describe, beforeEach, afterEach, it, expect, jest */` in the new test then and revert this if that's ok @rschallapalli?

jhevans (approved):

PR #631, opened 2022-07-21
PIC-2148: Refactor: Ungrouped content (Attendance table)

sarahyjja:


server/views/case-summary-record-order.njk
```diff
@@ -317,75 +317,92 @@
...
-            <div class="govuk-grid-column-one-third pac-!-grid-column-18 govuk-heading-m govuk-!-margin-bottom-0">
-                <h3 class="govuk-body govuk-!-font-weight-bold govuk-!-margin-bottom-0 govuk-!-display-inline-block">Acceptable</h3>
-            </div>
+        <div class="govuk-grid-row">
+            <dl class="govuk-grid-column-one-third pac-!-grid-column-18 govuk-!-margin-top-0">
+                <dt class="govuk-heading-m govuk-!-margin-bottom-0">
+                    <h3 class="govuk-body govuk-!-font-weight-bold govuk-!-margin-bottom-0 govuk-!-display-inline-block"
+                        aria-hidden="true">Acceptable</h3>
+                </dt>
+                <dd hidden="true">Acceptable</dd>
```
jhevans, line 335: I think we might have left this in accidentally 🤔 Can these elements be changed back to divs and spans and this line be removed because we've got the one at line 339?

server/views/case-summary-record-order.njk
```diff
@@ -317,75 +317,92 @@
...
-                <div class="govuk-grid-column-two-thirds govuk-!-margin-top-0 govuk-heading-m govuk-!-margin-bottom-1">
-                    <dt class="pac-dashboard-count">{{ attendedAppointments.failed | length }}</dt>
-                    <dd class="govuk-body govuk-!-font-weight-bold govuk-!-margin-left-0 govuk-!-margin-bottom-0 govuk-!-display-inline-block">Attendances</dd>
+        <div class="govuk-grid-row">
+            <dl class="govuk-grid-column-one-third pac-!-grid-column-18 govuk-!-margin-top-0">
+                <dt class="govuk-heading-m govuk-!-margin-bottom-0">
+                    <h3 class="govuk-body govuk-!-font-weight-bold govuk-!-margin-bottom-0 govuk-!-display-inline-block"
+                        aria-hidden="true">Unacceptable</h3>
+                </dt>
+                <dd hidden="true">Unacceptable</dd>
```
jhevans, line 377: Same as previous, don't think we need this line and the dl/dt tags can be divs/spans

jhevans (changes requested):

server/views/case-summary-record-order.njk
sarahyjja: yes you right!

jhevans (approved):
Good spot on the copy-pasted 'Acceptable' which I missed 😬

PR #659, opened 2022-09-06
Pic 2400 - case progress

rschallapalli:


public/stylesheets/app/_case-progress-card.scss
```diff
@@ -0,0 +1,99 @@
+.app-summary-card {
```
rschallapalli, line 1: Lift and shift from https://github.com/ministryofjustice/hmpps-manage-supervisions/blob/c08997614f346c9f331dc94cdbdb9fdcc3de25c0/src/client/sass/components/_summary-card.scss

server/views/court-case-progress.njk
```diff
@@ -0,0 +1,46 @@
...
+                        </div>
+                    </header>
+
+                    <div class="app-summary-card__body">
+                        <details class="govuk-details govuk-!-margin-bottom-0" data-module="govuk-details">
+                            <summary class="govuk-details__summary">
+                                <span class="govuk-details__summary-text">Add a note</span>
+                            </summary>
+                            <div class="govuk-details__text">
+                                <p class="govuk-!-margin-bottom-0">Coming soon...</p>
```
jhevans, line 28: @sarahyjja this is where the comments will go I think - @rschallapalli FYI @sarahyjja will pick up the front end work for this so it might be worth you two discussing any prior work you've done, or ideas you have about how we should go about that.

jhevans (approved):

PR #660, opened 2022-09-07
🤝 PIC-2420: Create contract for POST hearing to court-case-service

jhevans:
Proposed contract for POSTing and GETing hearing comments from `court-case-service`


tests/pact/case-service/create-hearing-comment.test.pact.js
```diff
@@ -0,0 +1,57 @@
...
+    })
+
+    it('Creates hearing comment successfully', async () => {
+      const addCommentResponse = {
+        ...requestBody,
+        created: '2022-08-24T15:42:41',
+        createdByUuid: 'cc2285f2-91b0-4e00-bd5e-9bdc35896bb6',
+        commentId: 12234
+      }
+      await provider.addInteraction({
```
jhevans, line 31: This is the contract

vb-moj (approved):

schemas/post-hearing-comment.schema.json
```diff
@@ -0,0 +1,51 @@
...
+      "type": "string",
+      "title": "The hearing id",
+      "default": ""
+    },
+    "createdByUuid": {
+      "$id": "#/properties/caseComments/items/anyOf/0/properties/createdByUuid",
+      "type": "string",
+      "title": "The created user uuid",
+      "default": ""
+    },
```
sarahyjja, line 36: I was wondering, would it be good to make the "createdByUuid" as a `required` parameter too?

rschallapalli (changes requested):
LGTM  (minor: there are some references to caseComments and caseId to be replaced with hearing) .. ignore me if this is still a draft

rschallapalli (commented):
In the hearing / case-progress area we are calling these as notes instead of comments so we may want to align the naming to notes?

schemas/post-hearing-comment.schema.json
jhevans: Fair shout, although @rschallapalli do we now need this at all? If it gets this from the user token then we can drop this entirely.

jhevans:
Yes good point - they're notes not comments! I will amend!

schemas/post-hearing-comment.schema.json
rschallapalli: No, we don't need any apart from author, comment, caseId

jhevans: Interesting - this schema defines the response, not the request. We're not checking that as such except for the Pact itself...

jhevans: It turns out that schema is serving as dual purpose for both the request and response with different fields appearing in each - I'll split them out to make the distinction clear

PR #664, opened 2022-09-13
PIC-2431: Allow rollout of specific functionality to individual court…

rschallapalli:
…s in Live

### DIFF
--- a/tests/utils/featureTogglePredicates.test.js
+++ b/tests/utils/featureTogglePredicates.test.js
@@ -17,11 +17,11 @@ describe('featureConfigPredicates', () => {
   })
 
   describe('enabledForUsers', () => {
-    it('should return true when a court is in the enabled list of courts', () => {
+    it('should return true when a user is in the enabled list of users', () => {
       expect(enabledForUsers('user-four', 'user-one').isEnabled({ username: 'user-one' })).toBe(true)
     })
 
-    it('should return false when a court is NOT in the enabled list of courts', () => {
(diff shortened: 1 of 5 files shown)


tests/utils/featureTogglePredicates.test.js
```diff
@@ -0,0 +1,116 @@
...
+      expect(enabledForCourts('B10AB', 'B10JQ').isEnabled({ court: 'B10JR' })).toBe(false)
+    })
+
+    it('should return false when null context is provided', () => {
+      expect(enabledForCourts('B10AB', 'B10JQ').isEnabled(undefined)).toBe(false)
+    })
+  })
+
+  describe('enabledForUsers', () => {
+    it('should return true when a court is in the enabled list of courts', () => {
```
jhevans, line 20:
```suggestion
    it('should return true when a user is in the enabled list of users', () => {
```

tests/utils/featureTogglePredicates.test.js
```diff
@@ -0,0 +1,116 @@
...
+      expect(enabledForUsers('user-four', 'user-one').isEnabled({ username: 'user-one' })).toBe(true)
+    })
+
+    it('should return false when a court is NOT in the enabled list of courts', () => {
```
jhevans, line 24:
```suggestion
    it('should return false when a user is NOT in the enabled list of users', () => {
```

jhevans (changes requested):

jhevans (approved):

PR #699, opened 2022-10-31
Pic 2016

rschallapalli:


rschallapalli (commented):
The size and position is way off
Cypress tests to be updated for the popup

rschallapalli:
Completed in a different PR

PR #707, opened 2022-12-01
PIC-2602: Exact search match on CRN

sarahyjja:


integration-tests/integration/features/case-list/case-list.spec.js
```diff
@@ -69,7 +69,7 @@ And('I should not see pagination', () => {
 })
 
 And('I should not see filters', () => {
-  cy.get('.pac-filter-card').should('not.exist')
+  cy.get('[data-test=pac-filter-card]').should('not.exist')
```
johnyfarrar, line 72: <3

johnyfarrar (approved):
LGTM

PR #727, opened 2023-03-10
Pic 2776

rschallapalli:


server/views/case-summary.njk
```diff
@@ -21,7 +21,7 @@
 {% block javascripts %}
     {{ super() }}
     <script type="text/javascript" src="/javascripts/case-progress-{{ appVersion }}.min.js"/>
-{% endblock %}
+{% endblock %}}
```
johnyfarrar, line 24: Is there an additional `}` there?

rschallapalli: Nope, good spot

rschallapalli: Removed 👍

PR #747, opened 2023-04-27
:chart_with_upwards_trend: Add Tracking for search

johnyfarrar:


server/routes/handlers/getCaseSearchHandler.js
```diff
@@ -1,9 +1,17 @@
 const log = require('../../../log')
+const trackEvent = require('../../utils/analytics')
 const getCaseSearchHandler = ({ searchCases }, getCaseSearchType) => async (req, res) => {
   const term = req.query.term
   const type = getCaseSearchType(term)
+  const { cookies } = req
   if (term && type) {
     const data = await searchCases(term, type)
+    trackEvent('CRN Search Performed', {
```
jhevans, line 9: Could we change the event name to be consistent with existing events which we've prefixed with `PiC` to make them easier to identify. An existing example is `PiCHearingUnchanged`

That way something like this will get you every event generated by any of our services:

```
customEvents
| where name startswith "PiC"
```

jhevans (changes requested):

server/routes/handlers/getCaseSearchHandler.js
johnyfarrar: Anything for you!

jhevans (approved):

PR #756, opened 2023-05-23
Pic 2964

rschallapalli:


public/javascripts/case-progress.js
```diff
@@ -3,18 +3,16 @@
...
+  function getAutoSaveHandler(textarea, url, inputDataFormatter) {
     let timeoutId;
-
     const noteEventListener = (event) => {
       // If a timer was already started, clear it.
       if (timeoutId) clearTimeout(timeoutId);
       // Set timer that will save comment when it fires.
       timeoutId = setTimeout(function () {
         // Make ajax call to save data.
         const xhr = new XMLHttpRequest()
```
paddybasi, line 14: just a curious point, but we can't use a js library I presume on the frontend? Also, I take it we can't support the new ```fetch``` method?

rschallapalli: We discussed among team about that but we're not sure we have used 3rd party libraries in MoJ and we were supposed to find that out but that didn't happen and I went with what other projects were doing

paddybasi: So I know in HMCTS (my previous team at MoJ) we used Angular (https://github.com/hmcts/rpx-xui-webapp)

public/javascripts/case-progress.js
```diff
@@ -35,11 +32,15 @@
     return noteEventListener
   }
 
-  const setupAutoSave = (textarea) => {
-    textarea.addEventListener('keypress', getAutoSaveHandler(textarea))
+  const setupAutoSaveHearingNote = (textarea) => {
+    textarea.addEventListener('keypress', getAutoSaveHandler(textarea,  'summary/auto-save-new-note', (textarea, event) => ({ note: event.srcElement.value, hearingId: textarea.dataset.hearingid }) ))
```
paddybasi, line 36: is ```keypress``` the correct event here? I know previously I've used the ```keyup``` event (which stops multiple events firing if someone has a key held down)

public/javascripts/case-progress.js
```diff
@@ -35,11 +32,15 @@
...
   }
 
-  document.querySelectorAll('.auto-save-text').forEach(setupAutoSave)
+  document.querySelectorAll('.auto-save-text').forEach(setupAutoSaveHearingNote)
```
paddybasi, line 39: this will suffer from the same issue we discussed previously, where you're binding the same handler multiple times (depending on how many ```.auto-save-text``` elements there are. Could you try using event delegation again?

paddybasi (changes requested):

public/javascripts/case-progress.js
rschallapalli: This bit has been test, delivered and is in prod so I don't want to disturb it now unless we agree with wider team on QA and doing UAT again. This is not part of the current ticket, I just had to change the function name to distinguish from autosave comment

public/javascripts/case-progress.js
rschallapalli: Updated

public/javascripts/case-progress.js
rschallapalli: Ah, that's great.

PR #758, opened 2023-05-26
Pic 2972 edit comment

rschallapalli:


integration-tests/integration/features/case-summary-comments/case-summary-comments.spec.js
```diff
@@ -35,20 +35,67 @@ And('I should see the following comments with the comment, author and date comme
...
+        }
       })
     })
   })
 })
 
-And('I click Delete on the below comment located in table row {int}', ($int, $data) => {
+const getCommentsRow = rowNum => cy.get('.govuk-table__body > .govuk-table__row').eq(rowNum - 1)
+
+And('I click {string} on the below comment located in table row {int}', ($link, $int, $data) => {
```
johnyfarrar, line 55: Really like making these more generic - good work :)

server/routes/handlers/getUpdateCommentRequestHandler.js
```diff
@@ -0,0 +1,9 @@
+const getUpdateCaseCommentHandler = caseService => async (req, res) => {
+  const { params: { courtCode, hearingId, defendantId }, body: { caseId, comment, commentId } } = req
+
+  await caseService.updateCaseComment(caseId, commentId, comment, res.locals.user.name)
```
johnyfarrar, line 4: Should there be a try catch / with an error state? Question not a requirement

johnyfarrar (commented):
A couple of comments - but I think it might be best to get this in? @paddybasi can you have a look too please?

server/routes/handlers/getUpdateCommentRequestHandler.js
rschallapalli: I have suggested something on that lines to handle any unexpected errors and show some graceful error message to the user on this thread but doing so is not inline with the recommendation. Please see this thread https://mojdt.slack.com/archives/C01FR4HKS3A/p1686821180541949

So the catch block has no clue what to do with the caught error.

paddybasi: yeah think there is a generic error handler that will catch any unknowns, but yes I think specific error handling could also be a thing to do

paddybasi (approved):
LGTM

server/routes/handlers/getUpdateCommentRequestHandler.js
johnyfarrar: Ok - I'll speak with the team about adding some work in the future about this.!

johnyfarrar (approved):

PR #770, opened 2023-06-21
PIC-2984: added outcomes page with filtering and sorting

paddybasi:


rschallapalli (changes requested):
LGTM minor changes needed

server/services/case-service.js
```diff
@@ -83,10 +84,63 @@ const createCaseService = (apiUrl) => {
       }
     },
 
+    getOutcomesList: async (courtCode, selectedFilterSorts, subsection) => {
+      const { filters, sorts } = getOutcomeListFilterSorts(selectedFilterSorts)
+
+      const filterMap = {
```
rschallapalli, line 90: Is it better to use value/label eg.  PROBATION_SENTENCE/'Probation sentence' in the dropdown options so that you don't have to do this mapping?

rschallapalli (changes requested):

server/services/case-service.js
```diff
@@ -83,10 +84,63 @@ const createCaseService = (apiUrl) => {
...
+
+      const paramMap = new URLSearchParams({
+        state: 'NEW'
+      })
+
+      filters.forEach(filter => filter.items.filter(item => item.checked).forEach(item => {
+        paramMap.append(filter.id, filterMap[item.value])
+      }))
+
+      const allowedSortValues = { ascending: 'ASC', descending: 'DESC' }
```
rschallapalli, line 108: Move this outside the method as a constant.

PR #772, opened 2023-06-28
Pic 3099 fix validation

rschallapalli:


johnyfarrar (approved):
Beautiful! Same with the other PR - will check on dev when ready!

PR #773, opened 2023-06-28
PIC-3102 fix back to cases link

rschallapalli:


johnyfarrar (approved):
LGTM - checked it locally against steps to reproduce in ticket. I'll Dev QA when it's in dev

PR #791, opened 2023-08-17
PIC-3261 fix duplicate court room filters

rschallapalli:


rschallapalli:
> Just looked this with @paddybasi - the filter manipulation wasn't the easiest to read but can see what it does - there might be some optimisation that could be done later but the tests look good - approved

happy to update if there are any pointers to do it better @johnyfarrar @paddybasi

rschallapalli:
> > Just looked this with @paddybasi - the filter manipulation wasn't the easiest to read but can see what it does - there might be some optimisation that could be done later but the tests look good - approved
> 
> happy to update if there are any pointers to do it better @johnyfarrar @paddybasi - practically we'll not improve it later - better do it now if it gives any performance gains

ok, later

PR #796, opened 2023-09-07
PIC-2883: in progress tab

paddybasi:
Added in progress tab

PR #798, opened 2023-09-13
Pic 2886 - assign outcome to user

johnyfarrar:


server/routes/index.js
```diff
@@ -582,6 +583,8 @@ module.exports = function Index ({ authenticationMiddleware }) {
 
   router.use('/:courtCode/outcomes', outcomesRouter)
 
+  router.post('/hearing/:hearingId/outcome/assign', defaults, catchErrors(assignHearingOutcomeRouteHandler))
```
rschallapalli, line 586: This can go to outcomes router here https://github.com/ministryofjustice/prepare-a-case/blob/e98ff8ed17b8e5089a67874e192ad2a13300acff/server/routes/handlers/outcomes/outcomesRouter.js#L16

server/routes/handlers/getAssignUserToOutcomeRequestHandler.js
```diff
@@ -0,0 +1,9 @@
+const getAssignUserToOutcomeRequestHandler = caseService => async (req, res) => {
+  const { params: { hearingId }, session, body: { targetDefendantId: defendantId, targetCourtCode: courtCode } } = req
+
+  await caseService.assignHearingOutcome(hearingId, res.locals.user.name)
+  session.assignHearingOutcomeSuccess = true
```
rschallapalli, line 5: This needs to be unset in the case summary handler, otherwise this will not be cleared and keeps rendering the notification on case summary page

public/javascripts/case-progress.js
```diff
@@ -143,11 +143,47 @@
...
+      const submitLink = '/hearing/' + target.dataset.hearingid + '/outcome/assign'
+      
+      targetDefendantIdInput.value = target.dataset.defendantid
+      targetCourtCodeInput.value = target.dataset.courtcode
+      assignOutcomeForm.setAttribute('action', submitLink)
+      assignOutcomeViewLink.href = viewLink
+      assignOutcomeModal.style.display = "block"
+    }
+
+    var assignLinks = document.getElementsByClassName('pac-assign')
```
paddybasi, line 179: So rather than assigning a handler (per row I presume here), could we have a single handler and use event propagation?

rschallapalli (changes requested):
Need some fixes

paddybasi (changes requested):
just a small request but looks good otherwise

public/javascripts/case-progress.js
johnyfarrar: You know that might just work - let me see if I have time to get that in!

server/routes/handlers/outcomes/outcomesRouter.js
```diff
@@ -14,5 +15,6 @@ const outcomesRouter = express.Router({ mergeParams: true })
 outcomesRouter.get('/', defaults, outcomesMiddleware('NEW'), catchErrors(casesToResultHandler))
 outcomesRouter.get('/in-progress', defaults, outcomesMiddleware('IN_PROGRESS'), catchErrors(casesInProgressHandler))
 outcomesRouter.get('/resulted-cases', defaults, outcomesMiddleware('RESULTED'), catchErrors(getResultedCasesHandler))
+outcomesRouter.post('/hearing/:hearingId/outcome/assign', defaults, catchErrors(assignHearingOutcomeRouteHandler))
```
rschallapalli, line 18:
```suggestion
outcomesRouter.post('/hearing/:hearingId/assign', defaults, catchErrors(assignHearingOutcomeRouteHandler))
```
?
outocme is already taken care in the  is already taken care of here https://github.com/ministryofjustice/prepare-a-case/blob/87e1c6833be200c89fe1976168c90d8b244d9f91/server/routes/index.js#L584

rschallapalli (changes requested):
One last

paddybasi (approved):
LGTM, just be aware that whichever PR gets merged in first (either this or https://github.com/ministryofjustice/prepare-a-case/pull/796 the tests on here may break)

rschallapalli (approved):

rschallapalli (approved):

paddybasi (approved):
LGTM, the only thing I think is that we should try making the modal more generic (so can use in other cases, like alerts etc if need be)

PR #809, opened 2023-09-29
Pic 3294 dynamic outcome count

rschallapalli:


integration-tests/integration/features/cases-in-progress-list-filter.feature
```diff
@@ -23,8 +25,6 @@ Feature: Cases In progress List filters
     And I click the "Outcome type" filter button
     And I click the "Apply filters" button
 
-#    Then I should see a tab with text "In progress (1)"
```
paddybasi, line 26: should we not put this back in to account for count when filtering?

paddybasi (dismissed):
LGTM, though I presume this is now under the assumption that the counts will be returned with the `STATE` api call (which also includes the updated counts taking into account the filters right)?

rschallapalli:
> LGTM, though I presume this is now under the assumption that the counts will be returned with the `STATE` api call (which also includes the updated counts taking into account the filters right)?

working on the filter part, the API doesn't take filters into account

server/views/outcomes/resultedCases.njk
```diff
@@ -1,4 +1,5 @@
 {%- set moment = params.moment -%}
+{%- set moment = params.moment -%}
```
johnyfarrar, line 2: Is that needed twice?

rschallapalli: Yea, plan B  😉 
accidental  CMD+D keys
will remove

johnyfarrar (approved):
LGTm

PR #814, opened 2023-10-09
Pic 3359 fix feature toggling

rschallapalli:


server/utils/features.js
```diff
@@ -17,7 +17,6 @@ const features = {
     ),
     allOf(
       isEnv('preprod'),
-      enabledForCourts(PLYMOUTH_MAGS_COURT_CODE, 'B05PK'),
```
paddybasi, line 20: is it no longer a requirement to have it enabled for plymouth?

paddybasi (dismissed):
LGTM

server/utils/features.js
rschallapalli: preprod - so removed court requirement to simplify granting access to the team

johnyfarrar (approved):
LGTM

PR #817, opened 2023-10-23
Pic 3395 miantenance page

rschallapalli:


johnyfarrar (approved):
LGTM

app.js
```diff
@@ -46,6 +46,20 @@ module.exports = function createApp ({ signInService, userService }) {
 
   app.set('view engine', 'njk')
 
+  if (config.maintenanceModeEnabled) {
+    app.use(express.static(path.join(__dirname, 'public'), { maxage: config.settings.assetCache }))
+
+    const router = express.Router()
+    const filePath = path.join(__dirname, 'public', 'maintenance.html')
+
+    router.use(async (req, res, next) => {
```
paddybasi, line 55: you've got an async call here which is probably not required?

app.js
```diff
@@ -46,6 +46,20 @@ module.exports = function createApp ({ signInService, userService }) {
...
+      res.sendFile(filePath)
```
paddybasi, line 56: Also I know it's a quick easy fix but I think we need to have some error handling in place too (just in case it can't read the file etc)

paddybasi (changes requested):

PR #821, opened 2023-11-02
Pic 3378

johnyfarrar:
Change the content on zero cases for outcomes workflow


server/views/outcomes/casesInProgress.njk
```diff
@@ -138,7 +138,10 @@
       {{ govukTable(tableData) }}
     {% endblock %}
   {% else %}
-    <p class="govuk-body">{{ "There are no cases that match these criteria." if params.filtersApplied else "No cases in progress available." }}</p>
+  {% if params.filtersApplied %}
+      <span class="govuk-heading-m">0 cases</span>
```
rschallapalli, line 142: Does the ticket mention this? I can see this on the ticket instead:

> 
>  If a user applies filters and there are no matching cases the content should read:
> 
> There are no matching cases.

server/views/outcomes/casesToResult.njk
```diff
@@ -120,7 +120,7 @@
     {% if params.filtersApplied %}
       <span class="govuk-heading-m">0 cases</span>
```
rschallapalli, line 121: Same here, is this expected? captured somewhere else?

rschallapalli (changes requested):

paddybasi (approved):
Aside from the comments @rschallapalli has mentioned it LTGM

server/views/outcomes/casesToResult.njk
johnyfarrar: Ticket has been updated!

PR #829, opened 2023-11-20
PIC-3245 prepping for redis upgrate

rschallapalli:


paddybasi (dismissed):
LGTM, the only thing on a separate note is that we shouldn't have to repeat ourselves with compose files https://docs.docker.com/compose/multiple-compose-files/extends/

ArronChave3 (approved):

PR #834, opened 2023-11-30
:chart_with_upwards_trend: Add Tracking for blank user name issue

johnyfarrar:
Adding more logging to the places we use the user.name property

### DIFF
--- a/server/routes/handlers/getAddCommentRequestHandler.js
+++ b/server/routes/handlers/getAddCommentRequestHandler.js
@@ -6,7 +6,15 @@ const getAddCaseCommentHandler = caseService => async (req, res) => {
   if (!comment) {
     session.caseCommentBlankError = true
   } else {
-    await caseService.addCaseComment(caseId, comment, res.locals.user.name)
+    try {
+      await caseService.addCaseComment(caseId, comment, res.locals.user.name)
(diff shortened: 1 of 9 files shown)


server/routes/handlers/getAddCommentRequestHandler.js
```diff
@@ -6,7 +6,15 @@ const getAddCaseCommentHandler = caseService => async (req, res) => {
   if (!comment) {
     session.caseCommentBlankError = true
   } else {
-    await caseService.addCaseComment(caseId, comment, res.locals.user.name)
+    try {
+      await caseService.addCaseComment(caseId, comment, res.locals.user.name)
+    } catch (err) {
```
rschallapalli, line 11: Suppressing the error here would render the case summary page but without the comment that the user has added on the case. Is that behaviour okay? 
Same for other catch blocks too?

Also all the handlers are wrapped with this function. Doesn't handling and logging the errors here suffice https://github.com/ministryofjustice/prepare-a-case/blob/b8bf89bc411e043f86956a73019f68a426daf72b/server/routes/handlers/catchAsyncErrors.js#L3-L20?

rschallapalli (approved):
LGTM

PR #846, opened 2024-01-08
Ac/dev tidy

andrewcharnley:
This very large commit is a collection of improvements that went down the rabbit hole with the aim of improving productivity for developers, reducing risk and enhancing performance.

**Improvements**

Mock services now map to real services.
Tests have many errors fixed.
Node is not required locally, only Docker.
Node version used in production is used in development (no sync needed).
All services inside Docker containers.
Minimised stdout noise in development.
Reduced 3rd party dependencies.
Probably other stuff I've forgotten about.
Performance enhancements

Screenshot 2024-01-16 at 21 49 12
Screenshot 2024-01-16 at 21 49 48

**Verification**

Unit and Int tests work in development and CI

**Notes**

PACT tests are broken and requires investigation. The skip flag was set on CI so it probably went unnoticed. The CI runner has been disabled.

Snyk (dependency checker) fails due to Axios wich is known to the developers and marked for removal. The CI runner has been disabled until this occurs.


.circleci/config.yml
```diff
@@ -45,8 +32,6 @@ commands:
           env_var: APP_VERSION
```
paddybasi, line 32: The above step `set_app_version` has been deleted, so where does this value come from now?

.circleci/config.yml
```diff
@@ -87,128 +65,88 @@ _snyk_options: &snyk_options
...
-          command: npm run unit-test
-      - store_test_results:
-          path: test-results
-      - store_artifacts:
-          path: coverage
+          name: Lint Test
+          command: docker run -it prepare-a-case-app /usr/src/app/node_modules/.bin/standard
+      - run:
+          name: Unit Test
+          command: docker run -e ENABLE_HEARING_OUTCOMES=false -it prepare-a-case-app /usr/src/app/node_modules/.bin/jest --config=./tests/jest.config.js --detectOpenHandles --ci --forceExit
```
paddybasi, line 96: is this correct? we've now got a dependency on a feature flag within the CI pipeline

.circleci/config.yml
```diff
@@ -87,128 +65,88 @@ _snyk_options: &snyk_options
...
       - run:
           name: Upload pact files to broker
-          command: |
-            npx pact-broker publish pact/pacts \
-              --broker-base-url $PACT_BROKER_BASE_URL \
-              --consumer-app-version $CIRCLE_SHA1 --tag $CIRCLE_BRANCH \
-              -u $PACT_BROKER_USERNAME -p $PACT_BROKER_PASSWORD
-
-  build_docker:
+          command: docker run -it prepare-a-case-app pact-broker publish pact/pacts --broker-base-url $PACT_BROKER_BASE_URL --consumer-app-version $CIRCLE_SHA1 --tag $CIRCLE_BRANCH -u $PACT_BROKER_USERNAME -p $PACT_BROKER_PASSWORD"
```
paddybasi, line 148: not sure running `pact-broker` in docker container by itself is valid, `npx` has been removed (unless you've added that as a dep in package.json?)

.gitignore
```diff
@@ -1,24 +1,18 @@
...
 .java-version
 process.yml
-yarn.lock
-yarn-error.log
-test-results/
 coverage/
-public/**/*.min.*
-public/stylesheets/*.css
-public/stylesheets/*.css.map
+public/build/javascripts/*.js
```
paddybasi, line 12: maybe just ignore everything in `build`? e.g. `/build/**`?

Dockerfile
```diff
@@ -1,38 +1,53 @@
...
+COPY ./schemas ./schemas
+## next statement used by PACT not wiremock!
+COPY ./wiremock ./wiremock
+ENTRYPOINT ["node"]
+
+## only use for development and integration tests
+FROM base as development
+RUN npm i -g concurrently
+ENTRYPOINT ["concurrently"]
+CMD [ "\"./node_modules/.bin/webpack --config ./public/config/webpack.config.js --watch\"", "\"./node_modules/.bin/nodemon --watch ./server ./bin/www\"", "\"./node_modules/.bin/nodemon --ext scss --watch ./public/src/stylesheets ./bin/build-css\"", "\"./node_modules/.bin/nodemon --ext js --watch ./public/src/javascripts ./bin/build-js\""]
```
paddybasi, line 53: tbh you can move all this compiling of js/scss into webpack itself and just get webpack to hot reload etc but sure that will bloat this PR further

.circleci/config.yml
andrewcharnley: Defo still there, line 68

README.md
```diff
@@ -17,93 +15,128 @@ Integration tests use [Cypress](https://www.cypress.io) and employ [Gherkin](htt
...
-env API_CLIENT_SECRET={client-secret} COURT_CASE_SERVICE_URL=http://court-case-service-dev.apps.live-1.cloud-platform.service.justice.gov.uk PAC_ENV=dev ENABLE_CASE_COMMENTS=true ENABLE_CASE_PROGRESS=true NOMIS_AUTH_URL=https://sign-in-dev.hmpps.service.justice.gov.uk/auth npm run start
+npm start
+npm run build
 ```
 
-Or to use alongside a local instance of *court-case-service* 
+With node being inside a docker container so to are the `node_modules` and if you make a change to package.json you will
+need to rebuild the docker container. Rebuilding the containers can be is fairly slow but you shouldn't need to do it day-to-day.
+
+Note that changes to wiremocks will not auto refresh, you'll need to restart the specific wiremock container.
```
paddybasi, line 39:
```
nodemon --watch mappings -e json --exec "curl -X POST http://localhost:9091/__admin/mappings/reset"
```
works for watching wiremocks and will reload into wiremock server

paddybasi (approved):
approved but ideally need some questions answered

.circleci/config.yml
andrewcharnley: Always was, if that feature flag isn't false the unit tests fail, so it's forced in CI.

.circleci/config.yml
andrewcharnley: Quite right but pact is currently disabled, it's in the commit message at the bottom.

Dockerfile
andrewcharnley: The feature js files are separates, I didn't want to start refactoring application code as part of this work.

README.md
andrewcharnley: It's a follow on piece of work. Feel free to do it!

I'm not sure if wiremock is node based, I think it's Java, in which case node won't be inside the official wiremock docker image.

PR #853, opened 2024-01-30
PIC-3270: Date Range Carousel

paddybasi:


Dockerfile
```diff
@@ -50,4 +50,4 @@ ENTRYPOINT ["node"]
 FROM base as development
 RUN npm i -g concurrently
 ENTRYPOINT ["concurrently"]
-CMD [ "\"./node_modules/.bin/webpack --config ./public/config/webpack.config.js --watch\"", "\"./node_modules/.bin/nodemon --watch ./server ./bin/www\"", "\"./node_modules/.bin/nodemon --ext scss --watch ./public/src/stylesheets ./bin/build-css\"", "\"./node_modules/.bin/nodemon --ext js --watch ./public/src/javascripts ./bin/build-js\""]
```
andrewcharnley, line 53: ???

docker-compose.yml
```diff
@@ -107,12 +113,13 @@ services:
...
     environment:
       - CYPRESS_FAIL_FAST_ENABLED=true
+      - CYPRESS_CASES_TOTAL_DAYS=${CASES_TOTAL_DAYS:-28}
       - DISPLAY=${DISPLAY}
     working_dir: /e2e
     command: ["open", "-C", "cypress/cypress.config.js"]
     volumes:
       #mac
-      #- /tmp/.X11-unix:/tmp/.X11-unix
+      - /tmp/.X11-unix:/tmp/.X11-unix
```
andrewcharnley, line 122: Pretty sure the cypress env isn't mac and this will break it?, perhaps.

package.json
```diff
@@ -10,7 +10,7 @@
     "url": "https://github.com/ministryofjustice/prepare-a-case.git"
   },
   "scripts": {
-    "start": "DISPLAY= docker compose up",
+    "start": "IP=$(ipconfig getifaddr en0) && /usr/X11/bin/xhost + $IP && DISPLAY=$IP:0 docker compose up",
```
andrewcharnley, line 13: Another that I'm unsure will work on non Mac. It isn't used on the CI though. Perhaps a start-mac or some type of script for setting it before it runs docker compose up?

public/config/external.js
```diff
@@ -1,2 +1,2 @@
 import 'bootstrap/js/dist/modal'
-// import 'bootstrap/dist/css/bootstrap.min.css'
```
andrewcharnley, line 2: Are you sure? I don't know why we're using it but @johnyfarrar may do.

Dockerfile
andrewcharnley: OK I see it moved to config, but CI builds it without the watch so you'd need to unwatch or the command will hang.

docker-compose.yml
paddybasi: this works for me in terms of exposing the cypress ui (and pipeline seems to be working fine?). I think we all have macs as devs by standard anyway

public/config/external.js
paddybasi: this was something I'd originally put in, it's not required as the css is being imported via scss files

server/views/case-list.njk
```diff
@@ -78,12 +78,26 @@
...
-{% set pastDate = moment().subtract(params.casesPastDays + 1, 'days') %}
-{% set useStartDate = pastDate if not isSunday else pastDate.subtract(1, 'days') %}
 {% set currentDate = moment(params.date, 'YYYY-MM-DD') %}
-{% set today = moment().format('YYYY-MM-DD') %}
+
+{% if currentDate.isBetween(moment().subtract(params.casesPastDays + 1, 'days'), moment().add(params.casesPastDays, 'days'), 'day', '[]') %}
+    {% set renderMiddleDate = moment() %}
+{% else %}
+
+    {# TODO: logic needs fixing here, but spent enough time on it already #}
```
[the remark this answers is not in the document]
paddybasi: fair enough, I was leaving it in to fix at a later date but I suppose we no longer support the 'non-js' version anyway, so this can be removed/simplified.

docker-compose.yml
paddybasi: FYI it also seems to work without it, so I've removed it

package.json
```diff
@@ -10,7 +10,9 @@
     "url": "https://github.com/ministryofjustice/prepare-a-case.git"
   },
   "scripts": {
-    "start": "IP=$(ipconfig getifaddr en0) && /usr/X11/bin/xhost + $IP && DISPLAY=$IP:0 docker compose up",
+    "start": "run-script-os",
```
andrewcharnley, line 13: I like it but run-script-os is going to run outside of docker, which defeats the point of the docker work to remove node being a stack requirement.

However given they will have some npm version to run these shortcuts, I would say use npx run-script-os and remove the entry from package.json. Defo don't want it in package.json

PS) unsure if start:detached will still work.

PPS) Update the readme with the mac xquartz thingy.

docker-compose.yml
```diff
@@ -119,7 +119,7 @@ services:
     command: ["open", "-C", "cypress/cypress.config.js"]
     volumes:
       #mac
-      - /tmp/.X11-unix:/tmp/.X11-unix
+#      - /tmp/.X11-unix:/tmp/.X11-unix
```
andrewcharnley, line 122: Are you 100% sure by rebuilding the container --build? I don't see how it can work if there isn't a socket.

FYI - On CI the integration tests run on a Linux VM now and I believe the Cypress default VM image is X11 not xorg (~/.Xauthority) based so this socket will be defined same as it is on Mac. Probably fine to leave it in, assuming CI passes.

public/src/javascripts/caselist-date-carousel.js
```diff
@@ -0,0 +1,35 @@
+import Splide from '@splidejs/splide'
+
+if (document.querySelector('section.splide')) {
```
andrewcharnley, line 3: Race condition, stick it inside

document.addEventListener('DOMContentLoaded', () => { ...

docker-compose.yml
paddybasi: So docker volumes will only get mounted at runtime, not at build time. I've tested it during runtime at it works fine without it. It's not needed in CI as this is only for displaying the cypress UI if we want it, shouldn't need it in CI pipeline

public/src/javascripts/caselist-date-carousel.js
paddybasi: yeah fair enough, I did originally have this in but thought it may not be needed as the external.js script is loaded at the very end? But just for completeness I'll put it back in

andrewcharnley (commented):
All good apart from one thing, can you stick those extra compose files into the compose folder. Trying to keep the root clean. Otherwise nothing else.

andrewcharnley (approved):

andrewcharnley (approved):

andrewcharnley (dismissed):

andrewcharnley (dismissed):
approved

ArronChave3 (approved):

johnyfarrar (approved):
LGTM

johnyfarrar (approved):
LGTM

andrewcharnley (approved):

PR #858, opened 2024-02-09
fix: corrects method for pointing local dev at cloud service

andrewcharnley:


ArronChave3 (approved):
How come we are changing port from 3007 to 3000?

andrewcharnley:
> How come we are changing port from 3007 to 3000?

It's because HMPPS Auth has allowlisted port 3000 and not 3007 and I couldn't be faffed with submitting a change request to them, easier to roll with what they expect.

PR #861, opened 2024-02-12
Feature/pic 3615

ArronChave3:


andrewcharnley (commented):
I'm on the fence so not approving or rejecting, the reason being is the filters don't change and requesting them upon every cases request is sub-optimal performance wise. Ideally I'd request them on service start and cache in memory, something like that.

There's also with certainty to be other pages that require this sort of 'meta' data, so maybe there could have been a more generic call rather than the case specific one.

ArronChave3:
> I'm on the fence so not approving or rejecting, the reason being is the filters don't change and requesting them upon each get cases request is sub-optimal, performance wise. Ideally I'd request them on service start and cache in memory, something like that.
> 
> There's also with certainty to be other pages that require this sort of 'meta' data, so maybe there could have been a more generic call rather than the cases specific one.
> 
> Not saying it's terrible though, different ways to skin the cat.

I had the same thought but wasn't sure if that would cause issues. How long would we have for cache expiry etc - happy to implement it but didn't want to cause any un needed complications

server/routes/handlers/outcomes/getResultedCasesHandler.js
```diff
@@ -19,7 +24,11 @@ const getResultedCasesHandler = caseService => async (req, res) => {
     items: prepareCourtRoomFilters(response.courtRoomFilters)
   }
 
-  const filters = [getOutcomeTypesListFilters(), courtRoomFilter]
+  // console.log('!!!!!!!!!filters', getOutcomeTypesListFilters)
+  console.log('!!!!!!!!!filters2', getOutcomeTypesListFilters())
```
andrewcharnley, line 28: :)

tests/routes/handlers/outcomes/getResultedCasesHandler.test.js
```diff
@@ -36,38 +55,47 @@ describe('getResultedCasesHandler', () => {
     }
     caseServiceMock.getOutcomesList.mockResolvedValueOnce(apiResponse)
 
+    console.log('--------------------case', caseServiceMock.getOutcomeTypes)
```
andrewcharnley, line 58: !

johnyfarrar (changes requested):

server/utils/getOutcomeTypesListFilters.js
```diff
@@ -1,14 +1,7 @@
...
-    { label: 'Adjourned', value: 'ADJOURNED' },
-    { label: 'Committed to Crown', value: 'COMMITTED_TO_CROWN' },
-    { label: 'Crown plus PSR', value: 'CROWN_PLUS_PSR' },
-    { label: 'No outcome', value: 'NO_OUTCOME' },
-    { label: 'Other', value: 'OTHER' }
-  ]
+const { getOutcomeTypes } = require('../../server/services/case-service')
+
+module.exports = async () => {
+  const outcomeTypes = await getOutcomeTypes()
```
rschallapalli, line 4: Can we not cache this instead of loading static data on every page load? 
```
var outcomeTypes
module.exports = async () => {
   outcomeTypes = outcomeTypes || await getOutcomeTypes()
```
could add some expiry logic. Or set in Redis with expiry. At least save it in session so that it is loaded once per user's session? Just worried if we are adding more overhead to already troubled CCS

andrewcharnley: This is what I said in the first post, the approach is T.D in the making.

johnyfarrar (approved):
LGTM

PR #865, opened 2024-02-19
Update Populate user to bust the cache if items are missing

johnyfarrar:


ArronChave3 (approved):

server/routes/middleware/populateCurrentUser.js
```diff
@@ -11,14 +11,18 @@ module.exports = userService => async (req, res, next) => {
         logger.info('Get user credentials from Redis cache.')
       }
       res.locals.user = { ...JSON.parse(getReply), ...res.locals.user }
-      return next()
+      if (!res.locals.user.name || !res.locals.user.userId) {
+        logger.warn('Error getting user credentials from Redis cache.', getReply)
```
rschallapalli, line 15: Can you log the user object as well here please? Just in case it is missing other details supposed to be populated from the JWT

PR #866, opened 2024-02-19
fix: wiremock lock version as 3.4.0 has breaking changes with templat…

andrewcharnley:
…ing that needs looking into


johnyfarrar (approved):
LGTM

PR #867, opened 2024-02-19
feat: PIC-3607 Outcome mass action (assign to)

andrewcharnley:
Filter fixed and into main. ready to proceed.


ArronChave3 (approved):
Something ive noticed is we don't have any setup for unit testing the client side javascript - this is definitely something we should do ASAP

PR #893, opened 2024-03-06
feat: PIC-3622 Workflow Prep Status Assign

andrewcharnley:


server/services/case-service.js
```diff
@@ -246,6 +246,17 @@ const createCaseService = apiUrl => {
       }
     },
 
+    workflow: {
+      tasks: {
+        state: {
+          set: (taskId, state, { hearingId, defendantId }) =>
+            update(`${apiUrl}/workflow/tasks/${taskId}/state?hearing=${hearingId}&defendant=${defendantId}`, {
```
rschallapalli, line 253: This is not restful 
PUT /hearing/{heairngId}/defendant/{defendantId}/prep-status/:state  please?

andrewcharnley: Prep is a task belonging to workflow. 

Defendant doesn't belong to hearing so I'm happily argue the original isn't restful.

rschallapalli: There is not resource called workflow and task in PaC. A defendant definitely belongs to a hearing. 
Either way I'd keep it consistent with existing practice unless we are talking about refactoring whole application as workflows and tasks

rschallapalli: @AndySweilam @johnyfarrar

andrewcharnley: No, a defendant belongs to a case.

There is context coming up for prep status which is workflow/task orientated and my approach is always to minimise new technical debt (even on projects drowning under it).

Later on there'll be something like this /workflow/${workId}/tasks/${taskId}/items/${itemId}/...

The other benefit of doing it this way is it's config driven, we don't need a hundred endpoint handlers all doing the same thing and we can react far quicker to change. ;)

rschallapalli: You can keep the node stuff to tackle the tech debt if you feel so but please adapt the URL to the one exposed by the backend.
PUT /hearing/{heairngId}/defendant/{defendantId}/prep-status/:state please

andrewcharnley: Unfortunately this wasn't resolved before Ravi left at the end of last week. The FE work was done prior to the BE and a conversation was missing inbetween (I was on holiday) and Ravi didn't look at the FE work prior to implementing.

Looking at workflow.tasks.state.set one can see this isn't task specific, but could be modified to use `/${taskId}-state/`, so long as any subsequent tasks use the same convention.

johnyfarrar (approved):
LGTM

ArronChave3 (approved):
LGTM

PR #894, opened 2024-03-06
Fix performance of bulk defendant matching

rschallapalli:


server/services/case-service.js
```diff
@@ -36,7 +36,20 @@ const createCaseService = apiUrl => {
       )) || { data: {} }
       return res.data
     },
-    getPagedCaseList: async (
+    getCaseListForBulkMatching: async (
```
rschallapalli, line 39: can be named appropriately. The endpoint should retrieve all the cases with Possible NDelius Record

rschallapalli:
https://github.com/ministryofjustice/prepare-a-case/pull/896 has been raised for thishence closed

PR #895, opened 2024-03-08
fix: CI pipeline to use docker-compose for lint and unit test

andrewcharnley:


.circleci/config.yml
```diff
@@ -80,26 +80,6 @@ jobs:
           value: "${VERSION}"
 
   test_lint_unit_pact:
-    executor: builder
```
markreesmoj, line 83: Have you seen the hmpps node executor as used here? https://github.com/ministryofjustice/manage-a-workforce-ui/blob/main/.circleci/config.yml#L22 
Possibly a simpler option than starting a docker instance to run some JavaScript tests and linting?

ArronChave3 (approved):

PR #898, opened 2024-03-14
Bug/pic 3711 unable to assign

ArronChave3:


server/services/case-service.js
```diff
@@ -413,6 +413,7 @@ const createCaseService = apiUrl => {
         author
       }),
     deleteCaseComment: async (caseId, defendantId, commentId) => {
+      //                         /cases/${caseId}/defendants/${defendantId}/comments/${commentId}
```
johnyfarrar, line 416: Just this comment to remove

johnyfarrar (commented):
Just a comment in there I think

PR #909, opened 2024-03-21
feat: Allow FE to target BE branch on local development and correct r…

andrewcharnley:
…elated issues

Note the court-case-service doesn't init the postgres schema (which it should) or auto-seed data on dev (which it should) meaning by default you'll hit a 500 o access.


compose/court-case-service.yml
```diff
@@ -7,38 +10,56 @@ services:
...
       hmpps-auth:
         condition: service_healthy
     ports:
       - "8080:8080"
+    entrypoint: ["/bin/sh", "/app/run.sh"]
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080/auth/health"]
     environment:
-      - SPRING_PROFILES_ACTIVE=dev
+      - SPRING_PROFILES_ACTIVE=postgres,local
```
Obsiye, line 23: just curious, does it need to have `postgres` here? can it just be `local`

Obsiye (approved):

compose/court-case-service.yml
Obsiye: no worries, it's not a blocking change . Approved the PR - LGTM.

PR #918, opened 2024-04-03
fix: if auth fires a 401/403 (should only be a 403) then the user is …

andrewcharnley:
…logged out and redirected to reauth. This replaces token refresh which as deprecated and removed (it never worked) from PAC.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -1,7 +1,5 @@
 const fs = require('fs')
-const axios = require('axios').default
 const { promisify } = require('util')
-const { Service } = require('axios-middleware')
 const config = require('./config')
 const express = require('express')
 const compression = require('compression')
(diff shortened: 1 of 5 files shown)


server/routes/middleware/authorisationMiddleware.js
```diff
@@ -8,7 +8,7 @@ module.exports = (req, res, next) => {
     const { authorities: roles, name, user_id: userId, user_uuid: uuid, user_name: username } = jwtDecode(res.locals.user.token)
     Object.assign(res.locals.user, { name, uuid, username, userId })
     if (!roles.includes(config.apis.oauth2.role)) {
-      log.warn(`User does not have required role ${config.apis.oauth2.role}`)
+      log.warn(`User does not have required role ${config.apis.oauth2.role}. Ensure Delius user has role 'CWBT200'.`)
```
andrewcharnley, line 11: Very quick 'heads up' for user, but rubbish code. Don't judge me on this one!

server/errorHandler.js
```diff
@@ -9,6 +9,14 @@ exports.notFound = (req, res, next) => {
   next(error)
 }
 
+exports.authError = (error, req, res, next) => {
+  if ([401,403]
```
andrewcharnley, line 13: I believe it will wire a 403 on expiry and the 401 is caught by the express auth firmware earlier on in the middleware (authorisation.js) and hence 401 is impossible, however I'm only 99.9% and can't test it until I see what auth does tomorrow (jwt expiry is now 12hrs). Technically it's assuming only the CCS can produce 40X, unsure if that's exactly the case being we use user-preference-service. Good enough for now.

server/routes/middleware/authorisationMiddleware.js
johnyfarrar: Nice!

server/app.js
```diff
@@ -201,57 +203,13 @@ module.exports = function createApp ({ signInService }) {
...
-        config.headers = {
-          ...config.headers,
-          ...axiosHeaders
-        }
-        return config
-      }
-    })
-    return next()
-  })
-
```
johnyfarrar, line 245: So this never worked? Is this where we used the middleware library? Can that go?

johnyfarrar (dismissed):
Couple of comments - LGTM

ArronChave3 (dismissed):

server/app.js
```diff
@@ -28,19 +28,29 @@ const hostName = os.hostname()
...
+  port: config.redis.port,
+  password: config.redis.password,
+  host: config.redis.host,
+  tls: config.redis.tls_enabled === 'true' ? {} : false
+})
+
+const service = new Service(axios)
+const app = express()
+
+if (!config.settings.reduceStdoutNoise) {
```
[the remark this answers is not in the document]
johnyfarrar: I think we add that as a ticket to fix forward?

server/app.js
andrewcharnley: No, two separate things. The passport express lib doesn't handle refresh functionality at all.

server/app.js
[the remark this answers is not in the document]
andrewcharnley: I agree Arron, it was to reduce the insane noise during dev as a first pass (turn it of and see if you want!). Not part of this work btw, previously implemented.

PR #919, opened 2024-04-03
PIC-3759: Stop filters persisting

ArronChave3:


tests/routes/view.test.js
```diff
@@ -748,6 +748,7 @@ describe('Routes', () => {
       return request(app)
         .get('/B14LO/outcomes?hearingDate=ASC&outcomeType=ADJOURNED')
         .then(response => {
+          console.log(response)
```
johnyfarrar, line 751: Console Log

server/views/case-list.njk
```diff
@@ -309,6 +311,9 @@
         {% endif %}
     {% endif -%}
 
+    hello
```
andrewcharnley, line 314: Hello Arron!

andrewcharnley (changes requested):
L314

server/views/case-list.njk
ArronChave3: hi

andrewcharnley (approved):

PR #932, opened 2024-04-11
[do not merge] PIC-2389: Update PACT tests for case comments feature

Obsiye:
## Update the Pact tests for case comments

- Update the endpoints for creating a case comment in the pact tests

This is because court case service and this service uses the endpoint `/cases/:id/defendants/:id/comments`
rather than `/cases/:id/comments`

This change should fix the PACT tests

**Additional information:**
This works on top of the previous PR
https://github.com/ministryofjustice/prepare-a-case/pull/658

## TO DO:
- [ ] Uncomment out pact_push from CI/CD
- [ ] Fix broken PACT tests or migrate to something else

### DIFF
--- a/schemas/post-case-comment-schema.schema.json
+++ b/schemas/post-case-comment-schema.schema.json
@@ -7,7 +7,8 @@
   "required": [
     "author",
     "comment",
-    "caseId"
+    "caseId",
+    "defendantId"
   ],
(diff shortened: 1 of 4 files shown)


Obsiye:
It hasn't been worked on since April. Closing.

PR #933, opened 2024-04-11
chore: convert helm to allowlist and tidy - probably breaks nginx whi…

andrewcharnley:
I think this will break the ingest.yaml but I'm not sure what it's trying to do here.


helm_deploy/prepare-a-case/values.yaml
```diff
@@ -0,0 +1,23 @@
+ingress:
```
markreesmoj, line 1: add `generic-service:` here

.circleci/config.yml
```diff
@@ -39,9 +39,8 @@ commands:
               --install --wait --reset-values --timeout 360s \
               --namespace=${KUBE_ENV_NAMESPACE} \
               --description="${APP_VERSION} - ${CIRCLE_BUILD_URL}" \
-              --values ~/git/helm_deploy/values-<< parameters.env >>.yaml \
-              --values - \
-              --values ~/git/helm_deploy/prepare-a-case/values-live.yaml \
+              --values ~/git/helm_deploy/prepare-a-case/values.yaml \
+              --values ~/git/helm_deploy/prepare-a-case/values-<< parameters.env >>.yaml \
```
markreesmoj, line 43: rather than modify this, there is a deploy_env job in the hmpps orb which will do the trick.

https://github.com/ministryofjustice/court-case-matcher/blob/main/.circleci/config.yml#L144

markreesmoj:
Add this to Chart.yaml:

```dependencies:
  - name: generic-service
    version: "3.8.0"
    repository: https://ministryofjustice.github.io/hmpps-helm-charts```

PR #943, opened 2024-04-23
PIC-3788: Initial page removal changes

ArronChave3:


wiremock/mappings/court-case-service/case-list/case-list-full.json
```diff
@@ -3,7 +3,7 @@
   "priority": 5,
   "request": {
     "method": "GET",
-    "urlPathPattern": "/court/([A-Z0-9]{5})/cases",
+    "urlPathPattern": "/court/([A-Z0-9]{5})/casesd",
```
andrewcharnley, line 6: Are you sure? :)

ArronChave3: I think that whole file just needs deleting actually

andrewcharnley (approved):

PR #949, opened 2024-05-20
Feature/pic 3835 refactor nunjucks templates

ArronChave3:
The idea going forward would be tot unit test all of the extracted nunchucks logic - this is why it is separated out into its own file. This approach means future code will be more robust

At the moment this is more of a "lift and shift" of the current logic - so there is room for many more improvements going forward


sameermoj (approved):
LGTM

sameermoj (approved):

PR #981, opened 2024-08-29
Apply filters from user preference service

ArronChave3:


server/routes/handlers/outcomes/getResultedCasesHandler.js
```diff
@@ -24,7 +24,7 @@ const getResultedCasesHandler = (caseService, userPreferenceService) => async (r
 
   const response = await caseService.getOutcomesList(
     courtCode,
-    req.query,
+    { ...filterParams, ...req.query },
```
johnyfarrar, line 27: I know it's not you - but can we use the `queryParams` variable like above? I know it's a sill thing. Just looks the same.

johnyfarrar (changes requested):
Pathetic feedback I know.

johnyfarrar (approved):
My hero!

PR #990, opened 2024-09-24
PIC-4093: Code change to display matching defendant records in descen…

sameermoj:
…ding order.

JIRA Link : https://dsdmoj.atlassian.net/browse/PIC-4092
JIRA Link : https://dsdmoj.atlassian.net/browse/PIC-4093

### DIFF
--- a/public/src/stylesheets/app/_overrides.scss
+++ b/public/src/stylesheets/app/_overrides.scss
@@ -29,6 +29,15 @@
   width: 45%;
 }
 
+.pac-\!-grid-column-sticky-45 {
+  width: 45%;
+  padding: 10px 0;
+  position: sticky;
+  top: 20px;
+  overflow: auto;
+  position: sticky;
+}
+
 .pac-\!-grid-column--max-700 {
   max-width: 700px;
 }
--- a/public/src/stylesheets/app/_overrides.scss
+++ b/public/src/stylesheets/app/_overrides.scss
@@ -29,15 +29,6 @@
   width: 45%;
 }
 
-.pac-\!-grid-column-sticky-45 {
-  width: 45%;
-  padding: 10px 0;
-  position: sticky;
-  top: 20px;
-  overflow: auto;
-  position: sticky;
-}
-
 .pac-\!-grid-column--max-700 {
   max-width: 700px;
 }
--- a/server/routes/handlers/getMatchingRecordRouteHandler.js
+++ b/server/routes/handlers/getMatchingRecordRouteHandler.js
@@ -23,7 +23,6 @@ const getMatchingRecordRouteHandler = (getMatchDetails, getCaseAndTemplateValues
     })
   }
 
-  // Parse page number and ensure it's a positive integer
   let currentPage = parseInt(queryParams.page, 10)
   if (isNaN(currentPage) || currentPage < 1) {
     currentPage = 1
@@ -31,11 +30,9 @@ const getMatchingRecordRouteHandler = (getMatchDetails, getCaseAndTemplateValues
 
   const recordsPerPage = settings.matchingRecordsToBeShownPerPage
 
-  // Calculate start and end index for slicing data
   const start = (currentPage - 1) * recordsPerPage
   const end = Math.min(start + recordsPerPage, offenderMatchDetails.length)
 
-  // Slice the data for the current page
   const paginatedMatchData = response.offenderMatchDetails.slice(start, end)
 
   const pageParams = {
--- a/server/routes/handlers/getMatchingRecordRouteHandler.js
+++ b/server/routes/handlers/getMatchingRecordRouteHandler.js
@@ -15,20 +15,11 @@ const getMatchingRecordRouteHandler = (getMatchDetails, getCaseAndTemplateValues
 
   const response = await getMatchDetails(defendantId)
 
-  // Check if offenderMatchDetails is an array
   const { offenderMatchDetails } = response
-  if (!Array.isArray(offenderMatchDetails)) {
-    return res.render('error', {
-      ...templateValues,
-      message: 'Unexpected data format received from the server.'
-    })
-  }
 
-  const matchingRecordsCount = offenderMatchDetails.length
-  if (matchingRecordsCount === 0) {
-    return res.render('no-match', {
-      ...templateValues,
-      message: 'No match details found for the defendant.'
+  if (!Array.isArray(offenderMatchDetails) || offenderMatchDetails.length === 0) {
+    return res.render('error', {
+      status: response.status || 500
     })
   }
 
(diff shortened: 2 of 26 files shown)


server/routes/index.js
```diff
@@ -656,22 +656,46 @@ module.exports = function Index ({ authenticationMiddleware }) {
...
-      } = templateValues
+
+      const { data: { defendantName } } = templateValues
       const response = await getMatchDetails(defendantId)
+
+      const matchingRecordsCount = response.offenderMatchDetails.length
+      const currentPage = parseInt(queryParams.page, 10) || 1
+      const limit = 5
+
+      // Calculate start and end index for slicing data
```
markreesmoj, line 675: Maybe this logic belongs in getPaginationObject? It seems to be derived from values we are passing into that method anyway?

johnyfarrar (changes requested):
Can we move all of this - 

Screenshot 2024-09-24 at 18 17 22

in to `server/routes/handlers/matching`

so that we don't have all that in the index.js

server/routes/index.js
```diff
@@ -656,22 +656,46 @@ module.exports = function Index ({ authenticationMiddleware }) {
...
       const templateValues = await getCaseAndTemplateValues(req)
       templateValues.title = 'Review possible NDelius records'
-      const {
-        data: { defendantName }
-      } = templateValues
+
+      const { data: { defendantName } } = templateValues
       const response = await getMatchDetails(defendantId)
+
+      const matchingRecordsCount = response.offenderMatchDetails.length
```
johnyfarrar, line 671: Do we ever check if this is zero? I know that to get here there must be a match as it's from a list but just wondering as it's a wild edge case.

johnyfarrar: Also as we're not typescript can we check`response.offenderMatchDetails` is an array...

server/services/case-service.js
```diff
@@ -34,6 +34,14 @@ const createCaseService = apiUrl => {
       const res = (await request(
         `${apiUrl}/defendant/${defendantId}/matchesDetail`
       )) || { data: {} }
+      // Sort offenderMatchDetails based on matchProbability in descending order
+      if (res.data.offenderMatchDetails) {
```
johnyfarrar, line 38: Do we want to check this is an array? just in case we get nothing?

server/routes/index.js
```diff
@@ -656,22 +656,46 @@ module.exports = function Index ({ authenticationMiddleware }) {
...
+      const currentPage = parseInt(queryParams.page, 10) || 1
```
johnyfarrar, line 672: This might be a bit over the top but what if someone changes the url and adds a minus or not a number?

server/routes/index.js
```diff
@@ -656,22 +656,46 @@ module.exports = function Index ({ authenticationMiddleware }) {
...
+      const limit = 5
```
johnyfarrar, line 673: Is the limit 50%?

server/routes/index.js
```diff
@@ -656,22 +656,67 @@ module.exports = function Index ({ authenticationMiddleware }) {
...
+
+      const { data: { defendantName } } = templateValues
       const response = await getMatchDetails(defendantId)
+
+      // Check if offenderMatchDetails is an array
+      const { offenderMatchDetails } = response
+      if (!Array.isArray(offenderMatchDetails)) {
+        return res.render('error', {
+          ...templateValues,
+          message: 'Unexpected data format received from the server.'
```
markreesmoj, line 676: Is there a test for this?

sameermoj:
In addition to these changes , one config change which might require test case fix is required.

server/routes/handlers/index.js
```diff
@@ -15,7 +15,7 @@ const {
   saveDraftCaseComment,
   deleteCaseCommentDraft,
   updateCaseComment,
-  updateHearingOutcomeToResulted
+  updateHearingOutcomeToResulted, getMatchDetails
```
johnyfarrar, line 18: new line...

server/views/match-defendant.njk
```diff
@@ -58,11 +58,11 @@
 
   <form name="matchDefendant" id="matchDefendant" method="post">
     <div class="govuk-grid-row">
-      <div class="govuk-grid-column-one-half pac-!-grid-column-45 govuk-!-padding-right-9">
+      <div class="govuk-grid-column-one-half pac-!-grid-column-sticky-45 govuk-!-padding-right-9">
```
johnyfarrar, line 61: Lovely!

public/src/stylesheets/app/_overrides.scss
```diff
@@ -29,6 +29,15 @@
   width: 45%;
 }
 
+.pac-\!-grid-column-sticky-45 {
```
johnyfarrar, line 32: Do we need to remove this?

server/routes/handlers/getMatchingRecordRouteHandler.js
```diff
@@ -0,0 +1,67 @@
...
+  const templateValues = await getCaseAndTemplateValues(req)
+  templateValues.title = 'Review possible NDelius records'
+
+  const { data: { defendantName } } = templateValues
+
+  const response = await getMatchDetails(defendantId)
+
+  const { offenderMatchDetails } = response
+
+  if (!Array.isArray(offenderMatchDetails) || offenderMatchDetails.length === 0) {
```
markreesmoj, line 20: Is there guaranteed to be at least one match? We are throwing an error if there are no matches, is that correct?

server/routes/handlers/getMatchingRecordRouteHandler.js
```diff
@@ -0,0 +1,67 @@
...
+    return res.render('error', {
+      status: response.status || 500
+    })
+  }
+
+  // Parse page number and ensure it's a positive integer
```
markreesmoj, line 26: Not sure how helpful these comments are, and there are a lot in this function - can we make the code more readable if we think the comments are necessary?

server/routes/handlers/getMatchingRecordRouteHandler.js
```diff
@@ -0,0 +1,67 @@
...
+  let currentPage = parseInt(queryParams.page, 10)
+  if (isNaN(currentPage) || currentPage < 1) {
+    currentPage = 1
+  }
+
+  const recordsPerPage = settings.matchingRecordsToBeShownPerPage
```
markreesmoj, line 32: Why is this a setting, but this is not?
`const maximumPages = 4`

server/routes/helpers.js
```diff
@@ -59,10 +59,76 @@ const prepareCourtRoomFilters = (allCourtRooms) => {
   return courtRooms
 }
 
+const getPaginationObject = (pageParams) => {
+  const maximumPages = 4
+
+  const { page: currentPage, matchingRecordsCount, recordsPerPage, courtCode, caseId, hearingId, defendantId } = pageParams
+
+  let startNum = Math.round(Math.ceil(currentPage - ((maximumPages - 1) / 2)))
```
markreesmoj, line 67: startNum is assigned here then potentially modified twice in the next 10 lines. Can we just assign it once? It might be easier to read the code that way
Same goes for endNum

server/routes/handlers/getMatchingRecordRouteHandler.js
sameermoj: So the flow of screen is :

Screenshot 2024-10-01 at 14 34 03


and then to 


Screenshot 2024-10-01 at 14 34 20


before coming to the below screen

Screenshot 2024-10-01 at 15 06 35

My assumption is that, by the time we are coming to this screen we have at least one match.

server/routes/handlers/getMatchingRecordRouteHandler.js
sameermoj: removed comments

server/routes/handlers/index.js
sameermoj: done

johnyfarrar (approved):
LGTM

PR #998, opened 2024-10-14
Add Plymouth to Pilot Courts

johnyfarrar:


server/utils/features.js
```diff
@@ -7,7 +7,7 @@ const LIVERPOOL_PRE_PILOT_USERS = ['TaylorColinoNPS', 'debbieleenps', 'emmacaddi
 
 const PLYMOUTH_MAGS_COURT_CODE = 'B50KH'
```
markreesmoj, line 8: Interesting that this code is already here

johnyfarrar: I didn't notice that.... made a change

PR #1002, opened 2024-10-28
Componentised and clean up filter component

ArronChave3:


server/views/case-list.njk
```diff
@@ -3,7 +3,7 @@
 {%- from "components/tabs/macro.njk" import pacTabs -%}
 
 {% block javascripts %}
-    <script type="text/javascript" src="/javascripts/filter-component-{{ appVersion }}.min.js"></script>
+    <script type="text/javascript" src="{{components.filterComponent.javascript}}"></script>
```
johnyfarrar, line 6: Is this a new way to get the file path? Not seen it and it looks cool?

ArronChave3: Ive moved the file path into the "nunchucksComponents" utils, the idea being to componentise the filter so we don't have to make changes in 4 places every time we update it

johnyfarrar (approved):
LGTM

PR #1005, opened 2024-11-08
Pic 4268 showing all possible records text

sameermoj:
JIRA Link : https://dsdmoj.atlassian.net/browse/PIC-4268

### DIFF
--- a/server/config.js
+++ b/server/config.js
@@ -42,6 +42,7 @@ module.exports = {
     scriptCache: '24h',
     matchingRecordsToBeShownPerPage: 5,
     maximumPages: 4,
+    minimumMatchProbability: 0.5,
     caseSearchResultPageSize: get('CASE_SEARCH_PAGE_SIZE', 20),
     hearingOutcomesPageSize: get('HEARING_OUTCOMES_PAGE_SIZE', 20),
     enableMatcherLogging: getBooleanParam('ENABLE_MATCHER_LOGGING'),
--- a/server/routes/helpers.js
+++ b/server/routes/helpers.js
@@ -103,10 +103,11 @@ const getPaginationObject = (pageParams) => {
     }
   }
 
+  const allMatchesShown = showAllMatches ? '&showAllMatches=True' : ''
   if (currentPage < totalPages) {
     nextLink = {
       text: 'Next',
-      href: `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${currentPage + 1}${showAllMatches ? '&showAllMatches=True' : ''}`
+      href: `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${currentPage + 1}${allMatchesShown}`
     }
   }
 
--- a/server/routes/helpers.js
+++ b/server/routes/helpers.js
@@ -63,7 +63,7 @@ const prepareCourtRoomFilters = (allCourtRooms) => {
 const getPaginationObject = (pageParams) => {
   const maximumPages = settings.maximumPages
 
-  const { page: currentPage, matchingRecordsCount, recordsPerPage, courtCode, caseId, hearingId, defendantId } = pageParams
+  const { page: currentPage, matchingRecordsCount, recordsPerPage, courtCode, caseId, hearingId, defendantId, showAllMatches } = pageParams
 
   let startNum = Math.round(Math.ceil(currentPage - ((maximumPages - 1) / 2)))
   let endNum = currentPage + Math.round(Math.ceil(((maximumPages - 1) / 2)))
@@ -88,9 +88,10 @@ const getPaginationObject = (pageParams) => {
   }
 
   for (let i = startNum; i <= endNum; i++) {
+    const href = `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${i}`
     pageItems.push({
       text: i,
-      href: `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${i}`,
+      href: showAllMatches ? `${href}&showAllMatches=True` : href,
       selected: currentPage === i
     })
   }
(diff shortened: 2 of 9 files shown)


server/config.js
```diff
@@ -42,6 +42,7 @@ module.exports = {
     scriptCache: '24h',
     matchingRecordsToBeShownPerPage: 5,
     maximumPages: 4,
+    minimumMatchProbability: 0.5,
```
markreesmoj, line 45: why has this come back? I thought it had just been removed

server/routes/helpers.js
```diff
@@ -88,24 +88,25 @@ const getPaginationObject = (pageParams) => {
...
+      href: `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${currentPage - 1}${showAllMatches ? '&showAllMatches=True' : ''}`
     }
   }
 
   if (currentPage < totalPages) {
     nextLink = {
       text: 'Next',
-      href: `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${currentPage + 1}`
+      href: `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${currentPage + 1}${showAllMatches ? '&showAllMatches=True' : ''}`
     }
```
markreesmoj, line 110: Can we just do this if condition once instead of 3 times?

server/config.js
sameermoj: In the ticket I missed the fact that this still needs to be there except that we remove this filtering when clicking on Show more records

server/services/case-service.js
```diff
@@ -31,25 +31,30 @@ const createCaseService = apiUrl => {
...
       if (Array.isArray(res.data.offenderMatchDetails)) {
         res.data.offenderMatchDetails.sort((a, b) => {
-          const probA = !isNaN(parseFloat(a.matchProbability))
-            ? parseFloat(a.matchProbability)
-            : 0
-          const probB = !isNaN(parseFloat(b.matchProbability))
-            ? parseFloat(b.matchProbability)
-            : 0
+          const probA = parseFloat(a.matchProbability ?? 0)
+          const probB = parseFloat(b.matchProbability ?? 0)
```
markreesmoj, line 47: is this just reverting the previous PR? I am confused by this

markreesmoj (commented):
❓

server/services/case-service.js
sameermoj: yes, to more extent it is reverting the previous change

sameermoj:
Its keeping the previous change and adding the Show more records link. Let me know if you want to have a chat. and I can explain.

tests/routes/view.test.js
```diff
@@ -513,7 +513,7 @@ describe('Routes', () => {
       '2e0afeb7-95d2-42f4-80e6-ccf96b282730'
     )
     expect(caseService.getMatchDetails).toHaveBeenCalledWith(
-      '2e0afeb7-95d2-42f4-80e6-ccf96b282730'
+      '2e0afeb7-95d2-42f4-80e6-ccf96b282730', undefined
```
markreesmoj, line 516: what is this `undefined`? is it something to do with the showAllMatches? If so could it be changed to a Boolean?

sameermoj: it is a second param that when not passed, defaults to undefined. This is in-place of showAllMatches.

server/routes/helpers.js
sameermoj: I did not get this comment tbh. It is being used where required

markreesmoj:
```
const allMatchesShown=showAllMatches ? '&showAllMatches=True' : ''
 if (currentPage < totalPages) {
    nextLink = {
      text: 'Next',
      href: `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${currentPage + 1}${allMatchesShown}`
    }
```

markreesmoj: It will be a bit easier to maintain that way. Really we should do the same with the href which is always the same apart from the page number

markreesmoj: The idea is not to have 3 checks for whether `showAllMatches` is `true`, all of which result in the same decision being made

markreesmoj (commented):
💬

server/routes/helpers.js
sameermoj: Thanks, have made the changes as per the above suggestion

tests/routes/view.test.js
markreesmoj: Is this the call which is being mocked here?
`const response = await getMatchDetails(defendantId, queryParams.showAllMatches)`

If so, it is followed a few lines later in the same method by 
`const showAllMatches = Boolean(queryParams.showAllMatches)`

How about moving `const showAllMatches` up a bit and passing it into `getMatchDetails` so that it's always either `true` or `false`?
That way you don't need to do `showAllMatches = Boolean(showAllMatches)` at the top of `getMatchDetails`

As far as I can see it is only called from server/routes/handlers/getMatchingRecordRouteHandler.js

sameermoj: have made the changes as per the above suggestion

server/services/case-service.js
```diff
@@ -31,25 +31,30 @@ const createCaseService = apiUrl => {
       const res = await request(`${apiUrl}/cases/${caseId}`)
       return res.data
     },
-    getMatchDetails: async defendantId => {
-      const res = await request(
+    getMatchDetails: async (defendantId, showAllMatches) => {
+      showAllMatches = Boolean(showAllMatches)
```
markreesmoj, line 35: this is not required - you have done this coercion before calling getMatchDetails

server/routes/helpers.js
```diff
@@ -88,24 +88,26 @@ const getPaginationObject = (pageParams) => {
...
 
   if (currentPage !== 1) {
     previousLink = {
       text: 'Previous',
-      href: `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${currentPage - 1}`
+      href: `/${courtCode}/case/${caseId}/hearing/${hearingId}/match/defendant/${defendantId}?page=${currentPage - 1}${showAllMatches ? '&showAllMatches=True' : ''}`
     }
   }
 
+  const allMatchesShown = showAllMatches ? '&showAllMatches=True' : ''
```
markreesmoj, line 106: Would be great to move this to line 89 and replace both other uses of `${showAllMatches ? '&showAllMatches=True' : ''}` with `$href$allMatchesShown` or similar

markreesmoj: The idea is to do the conditional check once only - we are still doing it 3 times. Once we know whether showAllMatches is true, we should not need to ask again for the purposes of building a URL

markreesmoj (approved):
💬

PR #1006, opened 2024-11-11
PIC-4115: Route handler for cancel link on match-defendant.njk

sameermoj:
JIRA Link : https://dsdmoj.atlassian.net/browse/PIC-4115


johnyfarrar (commented):
I have some comments and I'm not saying that I'm right in this -

Route handlers - 

Should we move to `caseSummary.js` rather than `getCaseSummary.js` 

Also I do think that the matcher route handlers should be a folder called `matcher`.

sameermoj:
Thanks @johnyfarrar , I will move the handlers to a separate matcher folder. 
I did not get the comment - Should we move to caseSummary.js rather than getCaseSummary.js

ArronChave3:
I would agree, I think all handlers should handle all methods (post/get etc) so for example, getCaseSummmary should just be renamed to caseSummary.js, and export a get function

sameermoj:
But there is no such file as getCaseSummary.js. I think what is being tried to refer as is to rename the new defined handlers eg: getMatchingRecordRouteHandler.js to matchingRecordRouteHandler.js and that too within matcher folder . Correct ? @johnyfarrar @ArronChave3

johnyfarrar:
> But there is no such file as getCaseSummary.js. I think what is being tried to refer as is to rename the new defined handlers eg: getMatchingRecordRouteHandler.js to matchingRecordRouteHandler.js and that too within matcher folder . Correct ? @johnyfarrar @ArronChave3 . And same for getCancelMatchRouteHandler.js and postDefendantMatchRouteHandler.js as well

Almost - they are in the routes folder so I would rename to `matchRecord.js`

sameermoj:
Based on the suggestions, have made changes, please have a look as and when possible. Thanks

ArronChave3 (approved):
LGTM

johnyfarrar (approved):
Great work!

PR #1007, opened 2024-11-15
PIC-4274: Adding logs and calling version2 of api for getting case list

sameermoj:
Changes for investigation : https://dsdmoj.atlassian.net/browse/PIC-4274

### DIFF
--- a/tests/services/case-service.test.js
+++ b/tests/services/case-service.test.js
@@ -158,7 +158,7 @@ describe('Case service', () => {
   })
 
   it('should call the API to request case list data', async () => {
-    moxios.stubRequest(`${apiUrl}/court/SHF/cases?date=2020-01-01`, {
+    moxios.stubRequest(`${apiUrl}/court/SHF/cases?date=2020-01-01&VERSION2=true`, {
       status: 200,
       response: {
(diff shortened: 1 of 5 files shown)


tests/services/case-service.test.js
```diff
@@ -158,7 +158,7 @@ describe('Case service', () => {
   })
 
   it('should call the API to request case list data', async () => {
-    moxios.stubRequest(`${apiUrl}/court/SHF/cases?date=2020-01-01`, {
```
Obsiye, line 161: Hi, just want to check if you are aware that the VERSION2 of that api paginates the results by default it returns 20 results. Is that ok?

Also the related court-case-service ticket is [here](https://dsdmoj.atlassian.net/browse/PIC-2377)

sameermoj: Hey thankd @Obsiye , this was discussed with @johnyfarrar and accordingly made changes

sameermoj: and yes, thats fine

Obsiye (approved):

johnyfarrar (approved):
LGTM

PR #1008, opened 2024-11-19
PIC-3156: Unlink nDelius page button/cancel link alignment

KiraNamida:
PIC-3156: Create a component for the GDS standard button input and cancel link along with int test steps.

image

### DIFF
--- a/server/views/components/button-group/template.njk
+++ b/server/views/components/button-group/template.njk
@@ -1,5 +1,8 @@
-<div class="govuk-button-group button-group">
-    <button class="govuk-button action-button">{{buttonText}}</button>
+
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+<div class="govuk-button-group page-button-group">
+    {{govukButton({text: buttonText, type: "submit"})}}
     {% if cancelLink %}
         <a href="{{cancelLink}}" class="govuk-link govuk-link--no-visited-state">Cancel</a>
     {% endif %}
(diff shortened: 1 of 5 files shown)


server/views/components/button-group/template.njk
```diff
@@ -0,0 +1,6 @@
+<div class="govuk-button-group button-group">
+    <button class="govuk-button action-button">{{buttonText}}</button>
```
Obsiye, line 2: i don't know this service well, sorry.

Is it ok that the button is defined a little differently [here](https://github.com/ministryofjustice/prepare-a-case/blob/main/server/views/components/button/template.njk) ? or should it use button-group too?

` <button id={{id}} type={{type}} class="action-button {{classes | join(' ')}}">{{text}}</button>`

KiraNamida: I would definitely like an institutional opinion on that. I went with the "minimum change required" approach in the first pass. Perhaps I should look at using that button inside this template to we still have one definition of a button component. 🤔

KiraNamida: Although, I have just dug for other references of that macro and it is only being imported in one file where it is not referenced and if I do use it, it throws a render error so something is going on here... 🙃

Obsiye: no worries, happy for you to ignore this comment 👍🏾

johnyfarrar (approved):
Yeah I like it! LGTM

PR #1013, opened 2024-11-21
PIC-3141: Capitalise case list acronyms

KiraNamida:
Capitalise acronyms to encourage them to be read as such, expand case-list tests to cover these badges.

This requires some follow up testing, particularly with JAWS which is not available on macOS but there is some follow up thinking about what to follow up anything else as "how acronyms are read" has as much to do with, if not more, with the screen readers settings.


johnyfarrar (approved):
LGTM

PR #1015, opened 2024-12-02
PIC 3136/text input improvements

KiraNamida:
[PIC-3136](https://dsdmoj.atlassian.net/browse/PIC-3136)

Improve a variety of accessibility points around our use of the accessible-autocomplete and the edit-courts page that it is used on.

- Upgrade the accessible-autocomplete package to v3.0.1 in order to make use of newer available parameters.
   - NOTE: This has caused npm to re-evaluate the package-lock.json config and reformat/upgrade/remove redundancies. This will require some regression testing to be sure we're happy with these changes but cleaning this up is on our todo list so presenting it here for consideration.
- Compile accessible-autocomplete css before govuk-frontend, this makes no difference to mutually exclusive bootstrap classes but allow the `govuk-input-error` to override the accessibile-autocomplete classes and match [Gov.UK Design System's error formatting](https://design-system.service.gov.uk/components/error-message/).
- Pass nonce down to the accessible-autocomplete wrapper component so we can run the enhancement script within it and directly manage parameters.
- Enhance accessibile-autocomplete wrapper component to provide complete suit of parameters to link up labels and correct error formatting.
   - Correct edit-courts error summary link to focus directly to the input, rather than the inline error message.
   - Add visually hidden text to Remove link to state which court it will remove when read in isolation.
   - Clean up button group to match Gov.UK styling.
   - Give the "Save button" the button role so that ATs treat it like a button as it is mimicking.

image

image

[PIC-3136]: https://dsdmoj.atlassian.net/browse/PIC-3136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "prepare-a-case",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
@@ -13,7 +13,7 @@
         "@microsoft/applicationinsights-web": "^3.3.2",
         "@ministryofjustice/frontend": "^2.1.0",
         "@splidejs/splide": "^4.1.4",
-        "accessible-autocomplete": "^2.0.4",
+        "accessible-autocomplete": "^3.0.1",
         "agentkeepalive": "^4.2.1",
         "applicationinsights": "2.9.2",
         "applicationinsights-native-metrics": "0.0.10",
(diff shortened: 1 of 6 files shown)


johnyfarrar (changes requested):
When we go in a touch a route in that horrible `index.js` can we move the handler in to it's own folder? I can show you an example?

KiraNamida:
Yeah, if there is a desire to distribute that into per route files (or at least small groups of related files) then doing it when the route is touched is no bad way to get on with making that happen.
I can have a look at that under here and push up a proposed reorganisation.

johnyfarrar (dismissed):
LGTM

johnyfarrar (approved):
LGTM

PR #1016, opened 2024-12-04
Spike: Possible route reoganisation

KiraNamida:
Possible route reorganisation: moving to a sub-route file and collecting with related handlers.

Whilst looking at PIC-3136 it was becoming apparent that the routes index file has become monolithic to the point where it is difficult to read, many handlers are inline and those handlers that are not are all in a single directory with little context.
As such we were asked to look at some reformatting to make this a more readable.

This PR represents the results of one such proposed solution*, the key here is that the individual features** are grouped into their own sub directory***, with a routes folder tasked with registering each of the related routes and the handlers contained in a further sub folder so they are easy to find next to the routes that make use of them.
This main route file would then call each of these in turn to register all of the routes.
This borrows heavily from common routing patterns found in the likes of react and other SPA applications where keeping related things together is sort after.
image

\* At the moment, the plan does not go so far as to propose changing all of these in one fell swoop, rather adopting a "if you work on a page, check it's routes and apply this is time/scope allows as we go".
This would mean the transition can be made is small changes rather than taking a huge chunk of resources.

** The term feature here is vague in it's application, this first example has been pulled together using the routes that begin `/my-courts` simple for the reason that the work that spawned was working on the set of pages that begin with `/my-courts`

*** This is also intended on a "if it makes sense", there are a number of catch all statements currently in the routes index that would remain, such as general error handling, authentication and health. These are unlikely to go anywhere as they are more global but we can see how things develop. Either way, it should eventually leave the index routes file as a much slimmer _jumping off point_ into more contained groups of routes.


server/routes/manage-courts/handlers/getUserSelectedCourts.js
```diff
@@ -1,5 +1,5 @@
-const { settings } = require('../../config')
-const trackEvent = require('../../utils/analytics')
+const { settings } = require('../../../config')
+const trackEvent = require('../../../utils/analytics')
 
 const getUserSelectedCourtsHandler = (getUserSelectedCourts) => async (req, res) => {
```
johnyfarrar, line 4: Just the naming on this one - I think that it should be `userSelectedCourts.js`

johnyfarrar:
The other thing is we have a pattern that we used before in the `routes/handlers/notifications` where we combined the Get and POST - do we want to use that?

johnyfarrar (approved):
Let's get this in!

PR #1021, opened 2024-12-10
Pic 3530 - move to aws secrets

johnyfarrar:


helm_deploy/prepare-a-case/templates/_envs.tpl
```diff
@@ -112,5 +112,17 @@ env:
...
+    valueFrom:
+      secretKeyRef:
+        name: liverpool-pre-pilot-users
+        key: liverpool_pre_pilot_users
+
+  - name: CASE_TRACKING_PRE_PILOT_USERS
+    valueFrom:
+      secretKeyRef:
+        name: case-tracking-pre-pilot-users
+        key: case_tracking_pre_pilot_users
```
markreesmoj, line 126: recommend moving to use the helm generic service ASAP in line with (most of) our other projects

markreesmoj: also pre-pilot :-)

markreesmoj (approved):
👍

helm_deploy/prepare-a-case/templates/_envs.tpl
johnyfarrar: Is that the key? Do I have a typo?

markreesmoj: no I just found the notion of a pre-pilot phase amusing.

PR #1022, opened 2024-12-10
PIC-3154: Risk register table column - non fixed width

KiraNamida:
Remove half width class from notes column on the risk register tables and allow them to auto set width, makes the most difference during zoom in/when the window is smaller (see the bottom most screencap).

Table before:
image
image

Table after:
image
image


sameermoj (approved):
LGTM

PR #1029, opened 2025-01-22
PIC 3143: Modal component, request UCD update

KiraNamida:
[[PIC-3143](https://dsdmoj.atlassian.net/browse/PIC-3143)]

A small change request by UCD, removing the left and right padding from the .modal-footer area so that the button align with the left hand content as based on the [Service Timeout](https://design.tax.service.gov.uk/hmrc-design-patterns/service-timeout/), which we are using as the design basis for our modals.

Left, before: footer has all around padding but notable on the left and right.
Right, after: footer still has padding on top and bottom but no longer on the left and right.
image

[PIC-3143]: https://dsdmoj.atlassian.net/browse/PIC-3143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/public/src/stylesheets/app/_bootstrap-modal.scss
+++ b/public/src/stylesheets/app/_bootstrap-modal.scss
@@ -173,7 +173,8 @@
   flex-wrap: wrap;
   align-items: center; // vertically center
   justify-content: flex-start; // put footer elements on the left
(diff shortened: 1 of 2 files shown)


johnyfarrar (approved):
LGTM

PR #1034, opened 2025-03-05
PIC-3232: Improve accessibility for case search errors

lmorison:
On case search error screen, adds "error" to title tag so screen readers read it out as soon as possible. Also ensures that error summary links to error field, complying with GDS guidelines.

Updates Cypress tests to assert the above.

### DIFF
--- a/server/views/case-search.njk
+++ b/server/views/case-search.njk
@@ -1,6 +1,6 @@
 {% extends "partials/layout.njk" %}
 
-{% set title = "Search result" %}
+{% set title = "Error: Search result" if searchError else "Search result" %}
 
 {% set tableData = {
     head: [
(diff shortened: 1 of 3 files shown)


server/views/case-search.njk
```diff
@@ -1,6 +1,6 @@
 {% extends "partials/layout.njk" %}
 
-{% set title = "Search result" %}
+{% set title = "Error: Search result" if searchError else "Search result" %}
```
Obsiye, line 3: Feel free to ignore the below:

you could do:
`set title = searchError ? "Error: Search result" : "Search result"`

Obsiye (approved):
LGTM thank you. i'd want to test this in dev/preprod

server/views/case-search.njk
lmorison: I don't think Nunjucks supports this syntax, happy to be corrected

Obsiye: ah thank you. i thought it was plain javascript

PR #1043, opened 2025-04-01
PIC-4550: use gds pagination

lmorison:
Use GDS pagination in PACFS

Before change:
image
image

After change:
image
image

PR #1045, opened 2025-04-28
use moj pagination in pac

lmorison:
Changes PAC back to use MOJ Pagination rather than GDS pagination which was introduced in https://github.com/ministryofjustice/prepare-a-case/pull/1043/files and deployed to dev but not pre-prod or prod

We are changing it back to MOJ pagination so that the 'results' section can be retained


Obsiye (approved):
LGTM 🚀

Obsiye (approved):

PR #1048, opened 2025-05-30
update axios version, remove axios-middlware, update affected code

lmorison:
Remove snyk scan and replace with recommended scans

Address security vulnerabilities highlighted by scans

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -212,19 +212,6 @@ workflows:
           main: true
           requires:
             - set_app_version
-      - hmpps/npm_security_audit:
-          slack_channel: << pipeline.parameters.alerts-slack-channel >>
-          context:
-            - hmpps-common-vars
-      - hmpps/trivy_latest_scan:
-          slack_channel: << pipeline.parameters.alerts-slack-channel >>
-          context:
-            - hmpps-common-vars
-      - hmpps/veracode_pipeline_scan:
-          slack_channel: << pipeline.parameters.alerts-slack-channel >>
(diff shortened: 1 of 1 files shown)


hmpps-user-preferences/index.js
```diff
@@ -41,11 +41,10 @@ app.put('/users/:userId/preferences/:preferenceName', (req, res) => {
                 console.error(err.message);
                 res.status(500).send('Internal server error');
             } else {
-                let putSql = 'INSERT INTO preference (hmpps_user_id, name, value) VALUES '
-                for (let i = 0; i < items.length; i++) {
-                    putSql += `('${userId}', '${preferenceName}', '${items[i]}')${i !== items.length - 1 ? ',' : ''}`
-                }
-                db.run(putSql, function (err) {
+                const placeholders = items.map(() => '(?, ?, ?)').join(', ')
```
Obsiye, line 44: hi, i just need help understanding what this does `items.map(() => '(?, ?, ?)').join(', ')` 

i'm assuming it is the (data1, data2, data3) when there are values. is this javascript destructuring going on here? my js knowledge is a little old

lmorison: hey, yeah that's right, and it serves as a placeholder for the values which will then be replaced using flatMap(). This is so that we can use SQLite's parameterised inputs as opposed to the previous approach which was being highlighted by security scans as being vulnerable to SQL injection (even though this is for mock of hmpps-user-preferences)

PR #1073, opened 2025-09-12
fix cancel button on match records page

lmorison:
The match records page expects defendant id and hearing id to be set, so have set these values accordingly

It was also reading court code from session rather than params (therefore it was undefined) so have resolved this too

Affects bulk matches and single case matches


SammyMarie14 (approved):
LGTM

PR #1140, opened 2026-03-26
Pic 5148/reapply cybersecurity issues

lmorison:
- Removed PACT tests (not ran for 2 years)
- Use generic-service helm template, and removed old helm code
- Updated values.yaml and values-env.yaml files to use generic-service
- Use generic-prometheus-alerts
- Removed unused env vars
- Fixed security vulnerabilities

Summarised by GitHub Copilot

**Summary: Prepare a Case — Infrastructure & Security Update**
This update is focused entirely on behind-the-scenes maintenance — there are no changes to how the application looks or behaves for end users.

**What's been done**
**Security vulnerabilities fixed** Several software dependencies (third-party code libraries the application relies on) had known security weaknesses. These have been updated to safer, more recent versions.

**Outdated test code removed** A set of automated tests called "PACT tests" had not been running for over two years and were no longer serving any purpose. These have been cleanly removed, reducing maintenance overhead.

**Deployment infrastructure modernised** The way the application is packaged and deployed to our hosting environments (dev, preprod, and prod) has been updated to use a shared, standardised approach used across MoJ digital services. This means:
- The deployment configuration is simpler and easier to maintain
- We benefit from improvements made centrally to that shared approach
- Future updates to deployment config will require less bespoke work

**Impact on users**
None in terms of functionality. However, there is a brief deployment consideration (see below).

⚠️ **Deployment action required**
The name of the ingress (the component that routes web traffic to the application) will change as part of this update. The existing ingress must be **manually deleted before deploying** to each environment, otherwise the deployment will fail. This is a one-time action per environment and is straightforward for the team to carry out — but it does mean the application will be **briefly unavailable** during each deployment window.

This should be planned and communicated in advance, particularly for the production deployment.

**Risk**
Medium. The deployment infrastructure change is significant in scope and requires the manual pre-deployment step above. A deployment to dev should be used to validate everything is working before proceeding to preprod and prod.


helm_deploy/prepare-a-case/values.yaml
```diff
@@ -0,0 +1,77 @@
+generic-service:
+  nameOverride: prepare-a-case
+  productId: 'HMPPS522'
+
+  serviceAccountName: "court-case-service"
+
+  image:
+    repository: quay.io/hmpps/prepare-a-case # todo: remove quay
```
lmorison, line 8: This will be done when the base image is changed in another ticket

Aron-v1 (dismissed):
LGTM

Aron-v1 (approved):
LGTM

leesykes (dismissed):

leesykes (approved):

PR #1151, opened 2026-05-13
ensure pagination works across all tabs

lmorison:
Pagination was not working on **Hearing outcome not required** or **Removed cases** tabs because it did not take into account the subsection when creating the pagination links

Resolved by passing the subsection when creating the base URL, allowing pagination logic to be tidied up

Added tests around this and refactored the test file to prevent test pollution

### DIFF
--- a/server/routes/handlers/getPagedCaseListRouteHandler.js
+++ b/server/routes/handlers/getPagedCaseListRouteHandler.js
@@ -19,8 +19,9 @@ const createBaseUrl = (params, queryParams) => {
   const questionMark = builtQueryParamString.length > 0
 
   const date = params.date ? params.date : getTodaysDate()
+  const subsection = params.subsection ? `/${params.subsection}` : ''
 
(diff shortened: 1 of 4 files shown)


server/routes/handlers/getPagedCaseListRouteHandler.js
```diff
@@ -19,8 +19,9 @@ const createBaseUrl = (params, queryParams) => {
   const questionMark = builtQueryParamString.length > 0
 
   const date = params.date ? params.date : getTodaysDate()
+  const subsection = params.subsection ? `/${params.subsection}` : ''
 
-  return `/${params.courtCode}/cases/${date}?${queryParamBuilder(remainder)}${questionMark ? '&' : ''}`
+  return `/${params.courtCode}/cases/${date}${subsection}?${builtQueryParamString}${questionMark ? '&' : ''}`
```
Aron-v1, line 24: This looks like it will do the job, but it's a little bit tricky to understand with building the url in this way. Could be worthwhile add a todo for a future refactor to use the URL object, and set the params that way.

Aron-v1 (approved):
LGTM. One comment to check.

PR #1156, opened 2026-06-12
Pic 5408 urn search

michaelpitt1:


tests/routes/handlers/caseSearchHandler.test.js
```diff
@@ -45,6 +49,36 @@ describe('caseSearchHandler', () => {
...
 
+  it('should invoke the service to search for cases by URN and render template', async () => {
+    const urnTerm = '01WW0298121'
+    const urnRequest = { query: { term: urnTerm, page: 1 }, cookies: { currentCourt: 'B12345' } }
+    const data = { items: [], totalElements: 3, totalPages: 1 }
+    getCaseSearchType.mockReturnValue({ searchType: 'URN' })
+    searchByCrnMock.mockResolvedValueOnce({ data })
+
+    await handler(urnRequest, mockResponse)
+    expect(searchByCrnMock).toHaveBeenCalledWith(urnTerm, 'URN', 1, 20)
```
Aron-v1, line 60: Double checking `searchByCrnMock` isn't a typo?

server/utils/getCaseSearchType.js
```diff
@@ -9,12 +9,18 @@ const getCaseSearchType = searchTerm => {
...
     }
     if (searchTermString.match(/\d/)) {
-      validations = searchTermString.match(/^[A-Za-z][0-9]{6}/) ? { searchType: 'CRN' } : { error: 'Enter a CRN in the format one letter followed by 6 numbers, for example A123456.' }
+      if (searchTermString.match(/^[A-Za-z][0-9]{6}/)) {
+        validations = { searchType: 'CRN' }
+      } else if (searchTermString.match(/^\d/)) {
+        validations = { searchType: 'URN' }
+      } else {
+        validations = { error: 'Enter a CRN in the format one letter followed by 6 numbers, for example A123456.' }
+      }
```
Aron-v1, line 18: On a practical note, are you handling leading or trailing whitespace?

**Less practical note:** 
This feels like an interim solution. I'm sure this pattern matching will work in some instances, but it doesn't seem robust. We are trying to make omni-search without a real omni search wired up on the BE. This is a product thing but it would make more sense to let the user explicitly search for things, maybe as part of an "advanced search", so that our users don't need to constantly go back into nDelius?

Aron-v1 (commented):
Two comments to check before approving.

Aron-v1 (approved):
LGTM.

PR #1171, opened 2026-07-29
add to regex so toggle outcome required doesn't break backlink

RichT900:


leesykes (approved):
Nice fix. And good tests coverage.

